PR #123, opened 2020-01-06
Traverse parent/child directories and search for a README source

tmickleydoyle:
Add directory walk to look for README files in case the user is not in the root directory or the README is located in a subdirectory.

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ var (
 	pager       bool
 	style       string
 	width       uint
+	walk		bool
 
 	rootCmd = &cobra.Command{
 		Use:           "glow SOURCE",
@@ -41,7 +42,7 @@ type Source struct {
 	URL    string
 }
 
-func readerFromArg(s string) (*Source, error) {
+func readerFromArg(cmd *cobra.Command, s string) (*Source, error) {
 	if s == "-" {
 		return &Source{reader: os.Stdin}, nil
 	}
@@ -92,16 +93,21 @@ func readerFromArg(s string) (*Source, error) {
 			}
 		}
 
-		filepath.Walk(s,
-			func(path string, info os.FileInfo, err error) error {
-			if strings.Contains(path, "README") {
-				fmt.Println(path)
-			}
+		// walk directory
+		if cmd.Flags().Changed("walk") {
+			filepath.Walk(s,
+				func(path string, info os.FileInfo, err error) error {
+				if strings.Contains(path, "README") {
+					fmt.Println(path)
+				}
 
-			return nil
-		})
+				return nil
+			})
 
-		return nil, errors.New("missing markdown source")
+			fmt.Println("Info: possible markdown sources")
+		}
+
+		return nil, errors.New("missing markdown source in current directory")
 	}
 
 	r, err := os.Open(s)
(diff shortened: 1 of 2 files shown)


tmickleydoyle:
Examples: ```./glow -k <directory>``` or ```./glow --walk <directory>```

meowgorithm:
🤔 I wonder if it should walk by default, seeing that by default it checks the current directory. Just a thought.

tmickleydoyle:
I did not want to change the default behavior of the base call `glow`. I am open 🎉

muesli:
A different feature, but slightly related:

I was thinking it should work the way around: similar to git, it could traverse through the parent directories, until it finds a valid README.

If you're in the sub-directory of a project's checkout then (say ~/src/foo/bar/zoo), it would still find the project's README, just like `git status` would still work in a sub-directory.

main.go
```diff
@@ -26,6 +26,7 @@ var (
 	pager       bool
 	style       string
 	width       uint
+	walk		bool
```
muesli: To avoid such situations, we agreed to the convention of using `gomt` or `goimports` to automatically format source code before committing anything. I can only recommend enabling that in your development environment of choice.

main.go
```diff
@@ -107,7 +122,7 @@ func execute(cmd *cobra.Command, args []string) error {
 	}
 
 	// create an io.Reader from the markdown source in cli-args
-	src, err := readerFromArg(arg)
+	src, err := readerFromArg(cmd, arg)
```
muesli: We probably need to split this up a bit. I'm not a fan of dragging the entire cmd into the little helper function. It also leads to other weird situations, where `-w` would be currently ignored if it's started with "github.com/foo/bar" as the source.

muesli:
After thinking about it a bit more, here's what I would suggest:

If `glow` gets started without a source, we behave like `git` and search the current and all its parent directories for a README.

If `glow` gets started with a specific source, we traverse that source should it be a directory. This means `glow .` would search the current dir and its children, and `glow foo` would search within `foo`.

That eliminates the need for a `-w` and any ambiguity that comes from trying to use it with HTTP/GitHub/GitLab sources.

main.go
tmickleydoyle: I have not worked with golang before. Do you have a development env setup that I could checkout? Thank you 🙇

muesli: I'm sure I can point you in the right direction! Which editor do you use?

main.go
```diff
@@ -41,7 +42,7 @@ type Source struct {
 	URL    string
 }
 
-func readerFromArg(s string) (*Source, error) {
+func readerFromArg(cmd *cobra.Command, s string) (*Source, error) {
```
tmickleydoyle:
```suggestion
func readerFromArg(cmd *cobra.Command, s) (*Source, error) {
```

tmickleydoyle:
I removed the list of all directories and sub-directories with a README getting printed out in favor of looking at the parent directory for a markdown file. I am using a `git` command to find the parent directory to prevent it from going back to the root of the machine. I reassign the path before it starts looking for the README files. Still working on this approach as I do not want to override is a user is trying to load a markdown file that is not a README.

```golang
	// walk to root
	if walkCmd {
		rootDir, _ := exec.Command(`git`, []string{"-C", s, "rev-parse", "--show-toplevel", "--sq"}...).Output()

		s = string(rootDir)
		s = strings.TrimSpace(s) + "/"
	}
```

srv1ssh:
Hi from CLI

PR #125, opened 2020-01-08
feat: support multiple files

caarlos0:
allow `glow *.md` to work by printing all args instead of just the first one :)

### DIFF
--- a/main.go
+++ b/main.go
@@ -101,11 +101,15 @@ func readerFromArg(s string) (*Source, error) {
 }
 
 func execute(cmd *cobra.Command, args []string) error {
-	var arg string
-	if len(args) > 0 {
-		arg = args[0]
+	for _, arg := range args {
+		if err := executeArg(cmd, arg); err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


muesli:
LGTM! Thank you @caarlos0, not just for this PR, but also for `goreleaser`, which means I can finally release stuff without losing my sanity :laughing:

muesli:
...and just as I click the merge button I realized this will actually break starting `glow` without any arguments. Don't worry though, I'm already fixing that.

caarlos0:
> ...and just as I click the merge button I realized this will actually break starting `glow` without any arguments. Don't worry though, I'm already fixing that.

ohhh, sorry about that 😞 

> LGTM! Thank you @caarlos0, not just for this PR, but also for `goreleaser`, which means I can finally release stuff without losing my sanity 😆

Thank you for all your work on many things as well :D

muesli:
> ohhh, sorry about that disappointed

I can't even blame you, I need to write tests for glow.

PR #147, opened 2020-05-14
Adding support for a GLAMOUR_STYLE environment variable

jessemillar:
A couple of CLI programs I use (e.g. https://github.com/sharkdp/bat#highlighting-theme) have support for defining the highlight/syntax theme via an environment variable. This PR adds similar functionality to Glow via a `GLAMOUR_STYLE` environment variable.

I propose the variable be named `GLAMOUR_STYLE` and any program using the Glamour library can honor the same variable to make configurations portable (I briefly entertained the idea of checking the environment variable in the `glamour` package instead of in Glow since that would automatically add the functionality to any projects that use `glamour` but I wasn’t sure if that was a good fit).

### DIFF
--- a/main.go
+++ b/main.go
@@ -183,7 +183,12 @@ func executeArg(cmd *cobra.Command, arg string, w io.Writer) error {
 	// initialize glamour
 	var gs glamour.TermRendererOption
 	if style == "auto" {
-		gs = glamour.WithAutoStyle()
+		envStyle := os.Getenv("GLAMOUR_STYLE")
+		if len(envStyle) > 0 {
+			gs = glamour.WithStylePath(envStyle)
+		} else {
+			gs = glamour.WithAutoStyle()
+		}
 	} else {
 		gs = glamour.WithStylePath(style)
 	}


muesli:
It's funny because I had a very similar implementation lingering around locally, but never pushed it, because I started questioning whether it belonged in `glow` or `glamour` myself. I even used the same name for the environment variable :laughing:

I'm still not fully convinced, but I think I came to the conclusion that it would be the most convenient to implement the env-var parsing in `glamour` itself, but make it an optional feature that users of the library can enable.

We reduce code duplication across projects, manage the handling, naming & conventions in a central place, but give developers the chance to not opt-in, should it break their existing app styling. For example I could imagine it being unwanted if `glamour` is used within an entirely themed & styled application.

What do you think?

jessemillar:
Ha! Glad we’re on the same page. With the new (to me) thought of env-var parsing being an opt-in feature of `glamour`, I think I’d definitely lean in that direction. Would an acceptable opt-in method be to introduce a new `.RenderWithConfig` (or better named) function to `glamour` that checks for `GLAMOUR_STYLE` and passes that to `.Render` overriding any manually passed styles?

jessemillar:
@muesli Just following up. Any thoughts on the separate function plan I proposed above? I could slap together a PR for `glamour` if you're okay with it.

muesli:
@jessemillar Apologies, I must have accidentally dismissed the notification about your previous comment. Yes, that would absolutely be acceptable!

jessemillar:
@muesli No worries! I'm not a big fan of the GitHub notifications interface. Glamour PR made [here](https://github.com/charmbracelet/glamour/pull/69). Closing this PR in favor of that one.

PR #151, opened 2020-07-17
Generate new URL string for each entry in `readmeNames`

### LINKED ISSUE #134: Make it aware of lowercase readme by default
Make runner aware of `readme.md` as well. Some people prefer all files to be lowercased, including readme. Now it's required to pass such name explicitly, wasting time.

angadgill92:
The code in `github.go` and `gitlab.go` mutates the same URL path
making anything but the first attempt incorrect. This means the project
at the moment only works for repositories that have `README.md` present.

I added a debug log to `findGithubREADME` and tried to get the readme
of one of my own projects (it has a readme.md)

here is the output:

```shell
$ go run . github.com/angadgill92/clean
TRYING PATH  /angadgill92/clean/master/README.md
TRYING PATH  /angadgill92/clean/master/README.md/master/README
Error: can't find README in GitHub repository
Usage:
  glow SOURCE [flags]

Flags:
  -h, --help           help for glow
  -p, --pager          display with pager
  -s, --style string   style name or JSON path (default "auto")
      --version        version for glow
  -w, --width uint     word-wrap at width

exit status 255
```

The reason for this is that we're doing
```go
   v := u
   v.Path += "/master/" + r
```

and on every iteration of the loop it adds `"/master/" + r` to
whatever URL was generated in the last iteration. This makes
every URL generated by `glow` after the first one invalid causing
the error.

Also fixes #134 🙂

### DIFF
--- a/github.go
+++ b/github.go
@@ -28,18 +28,17 @@ func findGitHubREADME(s string) (*source, error) {
 		return nil, err
 	}
 	u.Host = "raw.githubusercontent.com"
-
+	u.Path += "/master/"
 	for _, r := range readmeNames {
-		v := u
-		v.Path += "/master/" + r
+		v := u.String() + r
 
-		resp, err := http.Get(v.String())
+		resp, err := http.Get(v)
 		if err != nil {
 			return nil, err
 		}
 
 		if resp.StatusCode == http.StatusOK {
-			return &source{resp.Body, v.String()}, nil
+			return &source{resp.Body, v}, nil
 		}
 	}
 
--- a/gitlab.go
+++ b/gitlab.go
@@ -27,18 +27,17 @@ func findGitLabREADME(s string) (*source, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	u.Path += "/raw/master/"
 	for _, r := range readmeNames {
-		v := u
-		v.Path += "/raw/master/" + r
+		v := u.String() + r
 
-		resp, err := http.Get(v.String())
+		resp, err := http.Get(v)
 		if err != nil {
 			return nil, err
 		}
 
 		if resp.StatusCode == http.StatusOK {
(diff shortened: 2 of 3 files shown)


angadgill92:
Hi, just checking if there are any plans to merge this? If there are any issues please let me know? 🙏

muesli:
Hey @angadgill92! Sorry for the late response. You're certainly right, we should do better here. I wonder if there's a way to find a file (entirely) case-insensitively - like we do for local files - but I fear there's no way that doesn't involve the GitHub or GitLab APIs. Otherwise the next corner-case would be a `Readme.md`, then a `readme.MD`, and so on...

angadgill92:
Hey, no worries 🙂 You make a valid point, I wonder if there's a way to list the contents of the root directory of a GitHub/Gitlab repo, without cloning it 🤔. If we could get a list of root directories via some API (as you rightly pointed out), we could pattern match the paths case insensitively and pick the correct path for the readme, and then make a single call to fetch it instead of trying to brute force.

Are you apprehensive about using GitHub API/Gitlab API? If it's okay with you I'd like to find some time over this weekend to explore this, and update the PR.

muesli:
Since we already have GitHub / GitLab specific implementations I'm not really opposed, I was just hoping to find something a bit more generic, that would also support the likes of a self-hosted Gitea (or similar). I don't think that's easily possible though.

Definitely looking forward to what you come up with!

cristiand391:
Hi, is someone already working on this?

I have a very simple implementation using the [github api](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-a-repository-readme) and would like to adapt it to glow if possible.

muesli:
By all means @cristiand391, go ahead. Curious to see what you come up with!

caarlos0:
I think this was fixed on master already, #450 


thanks, and sorry for the delay

PR #152, opened 2020-07-23
Enable use of the `GLAMOUR_STYLE` environment variable

jessemillar:
This PR uses `v0.2.0` of Glamour to enable the `GLAMOUR_STYLE` environment variable for consistent color theme configs.

The way the PR is currently written prioritizes `GLAMOUR_STYLE` over the `--style` flag. I'm not a big fan of that (since `--style` seems more explicit and should likely override an environment config). I'm stumped on an `if` logic flow that prioritizes `--style` first though since `style` gets set to `notty` if `--style` is not explicitly passed (and we likely want to respect `notty`). Any thoughts?


main.go
```diff
@@ -182,7 +182,9 @@ func executeArg(cmd *cobra.Command, arg string, w io.Writer) error {
 
 	// initialize glamour
 	var gs glamour.TermRendererOption
-	if style == "auto" {
+	if len(os.Getenv("GLAMOUR_STYLE")) > 0 {
```
muesli, line 185: I think we can skip this check. If `style` equals `auto` we can always use `WithEnvironmentConfig`, as it will also default to `auto` if no environment variable was set. In other cases we would defer to `WithStylePath`.

jessemillar: Works for me. Updated.

muesli:
Thank you for following up on your earlier contribution! Much appreciated!

PR #160, opened 2020-10-03
Use GitHub/GitLab API to get a repo readme

cristiand391:
Ref: https://github.com/charmbracelet/glow/pull/151#issuecomment-700212676

This PR adds initial support for using the GitHub/GitLab public API to get a repo readme file.

<s>
TODO:

Rate limiting:
The GitHub API [allows up to 60 requests per hour](https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#rate-limiting), in the case the user exceeds the limit, how should Glow handle it?

GitLab [allows up to 10 req per second](https://docs.gitlab.com/ee/user/gitlab_com/#haproxy-api-throttle) so it shouldn't be a problem.
</s>

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -1,11 +1,20 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
 )
+
+// gitLabRepo contains information about a GitLab repo
+type gitLabRepo struct {
+	Branch    string `json:"default_branch"`
+	ReadmeURL string `json:"readme_url"`
+}
 
 // isGitLabURL tests a string to determine if it is a well-structured GitLab URL
 func isGitLabURL(s string) (string, bool) {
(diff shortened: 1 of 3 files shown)


muesli:
Thanks for your PR, @cristiand391! I'll try this out and review it asap!

gitlab.go
```diff
@@ -22,24 +31,44 @@ func isGitLabURL(s string) (string, bool) {
...
 // findGitLabREADME tries to find the correct README filename in a repository
-func findGitLabREADME(s string) (*source, error) {
-	u, err := url.ParseRequestURI(s)
+func findGitLabREADME(repoURL string) (*source, error) {
+	user, repo, err := userAndRepoFromURL(repoURL)
+	if err != nil {
+		return nil, err
+	}
+
+	apiRepoURL := fmt.Sprintf("https://gitlab.com/api/v4/projects/%s%s%s", user, "%2F", repo)
```
YerinAlexey, line 40: To be safe, this should probably be escaped.

Ref: [`url.PathEscape`](https://godocs.io/net/url#PathEscape)

```suggestion
	apiRepoURL := fmt.Sprintf("https://gitlab.com/api/v4/projects/%s", url.PathEscape(user + "/" + repo))
```

gitlab.go
```diff
@@ -22,24 +31,44 @@ func isGitLabURL(s string) (string, bool) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	for _, r := range readmeNames {
-		v := u
-		v.Path += "/raw/master/" + r
+	repoInfo := gitLabRepo{}
+
+	if err := json.Unmarshal([]byte(string(content)), &repoInfo); err != nil {
```
YerinAlexey, line 54: This is useless conversion as `ioutil.ReadAll` already returns `[]byte`.

Ref: [`ioutil.ReadAll`](https://godocs.io/io/ioutil#ReadAll)

```suggestion
	if err := json.Unmarshal(content, &repoInfo); err != nil {
```

utils.go
```diff
@@ -0,0 +1,22 @@
...
+	"strings"
+)
+
+func userAndRepoFromURL(repoURL string) (string, string, error) {
+	u, err := url.ParseRequestURI(repoURL)
+	if err != nil {
+		return "", "", err
+	}
+
+	pathSplit := strings.Split(u.Path, "/")
```
YerinAlexey, line 15: I think it is better to use [`path.Split`](https://godocs.io/path#Split) here as it will work on non-Unix systems as well.

YerinAlexey (changes requested):

PR #163, opened 2020-10-09
Remove duplicate "down/j" from the "?" help

bubersson:
"down/j" was there twice.


muesli:
Oh wow, embarrassing! Thanks for the fix! :smiley:

PR #186, opened 2020-10-20
Now in Scoop!

batkiz:
[scoop](https://github.com/lukesampson/scoop) is a command-line installer for Windows.

by https://github.com/ScoopInstaller/Main/pull/1510, now we could install glow with scoop!


muesli:
Thank you @batkiz!

Issue #182, opened 2020-10-19
Open in editor


runar-indico:
When viewing a file with less, one can press `v` to open the file in the editor. Adding this feature to glow would be great.

lepasq:
Hi! I'd like to work on this issue.

This shouldn't take too long, so I'll try to make a PR within the next days.

muesli:
@lepasq Awesome, looking forward to that!

lepasq:
I've worked on the feature for a bit, and so far I encountered one problem: 

Executing any editor will crash glow right after exiting the editor.
You can try this yourself by running `exec nvim` or `exec nano` in your terminal emulator, and quitting nvim/nano.

I couldn't find a workaround to this issue yet.
A StackOverflow user explained the problem [here](https://stackoverflow.com/a/63514108).

Shall I make a PR in the meantime, such that others can have a look at my code and try to solve it?


![Preview Gif](https://media.giphy.com/media/IslgRUjFCGek7GmTxe/giphy.gif)
Error message in terminal: `Error: read /dev/stdin: resource temporarily unavailable`

meowgorithm:
@lepasq Yes, please do! You could also post a link to your fork as well if you prefer.

lepasq:
@meowgorithm I just pushed my changes to the `pager-edit-key` branch of my [fork](https://github.com/lepasq/glow/tree/pager-edit-key) in this [commit](https://github.com/lepasq/glow/commit/ae56f7a25b0114c0fc98b72b958e77822408ffa4).

lepasq:
Note that the code won't work on Windows yet, since I'm running the editor by launching its binary in `/bin`.

wantyapps:
You can check the value of the "$EDITOR" variable, and launch it.

lepasq:
> You can check the value of the "$EDITOR" variable, and launch it.

Fair point. This is what the method currently looks like:
```go
func findEditor() (string, error) {
	if os.Getenv("EDITOR") != "" {
		return os.Getenv("EDITOR"), nil
	}
	editors := []string{"nvim", "vim", "vi", "nano", "gedit"}
	for _, editor := range editors {
		if _, err := exec.LookPath("/bin/" + editor); err == nil {
			return "/bin/" + editor, nil
		}
	}
	return "", fmt.Errorf("Couldn't find an editor on this system.")
}
```
Do we want to remove the second part, in which we check if the user has an editor inside `/bin`, in case that the user did not set the environment variable?  
[...]

PR #214, opened 2020-12-03
Added option to edit current document

shitchell:
Implements charmbracelet/glow#182

Allows the user to edit the current document by typing 'e'. Glow will search $PATH for a list of pre-defined command line text editors, giving priority to the command specified with $EDITOR if set. A temporary file is created with the contents of the current document, and the chosen text editor is used to edit it. If no changes are made when the editor exits, nothing happens. If changes *are* made, then:

**if** the document is a local file, the edits are written to that file
**if** the document is stashed, a new stash is created with the edits, and the old stash is deleted.

It would be ideal to simply update the contents of a stash, but the charm API doesn't currently allow this. Deleting the selected stash has the side effect that, when the user returns to the list of files and tries to re-open the same document, the stash won't load. Restarting `glow` solves this and shows the newly created, updated stash.

https://github.com/charmbracelet/charm/pull/13 creates a `SetMarkdownBody()` method to resolve this minor issue.

This PR also adds a few status messages throughout the process. To that end, `showStatusMessage()` was updated to work as it was presumably intended. At present, it only ever displays "Stashed!" no matter what parameter is sent. This PR updates it so that you can display any message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 glow
 dist/
+.envrc


Siilwyn:
Awesome! I often find myself opening the file with an editor to make some small adjustments. First time using glow I tried hitting `e` so this seems an intuitive addition too.

wantyapps:
Finally!

ui/pager.go
```diff
@@ -199,6 +202,92 @@ func (m pagerModel) Update(msg tea.Msg) (pagerModel, tea.Cmd) {
...
+				cmd.Stdin = os.Stdin
+				cmd.Stdout = os.Stdout
+				cmd.Stderr = os.Stderr
+				err = cmd.Start()
+				if err != nil {
+					log.Println(err)
+				}
+				err = cmd.Wait()
+				if err != nil {
+					log.Println(err)
```
meowgorithm, line 242: There's a package-level variable `debug bool` which, when true, indicates that logs will go to a file. That variable should be checked prior to logging, otherwise logging to stdout will either mess with the UI or simply not be shown. To log to a file you can set `GLOW_LOGFILE`, by the way.

Beyond that, however, we should be properly handling errors in the UI rather than logging them.

ui/pager.go
```diff
@@ -199,6 +202,92 @@ func (m pagerModel) Update(msg tea.Msg) (pagerModel, tea.Cmd) {
...
+					cmds = append(cmds, cmd)
+					break
+				}
+
+				// Determine if the current file is stashed or not
+				isStashed := m.currentDocument.markdownType == stashedMarkdown ||
+					m.currentDocument.markdownType == convertedMarkdown
+				
+				// Copy the contents of the document to a temporary file
+				tempFile, err := ioutil.TempFile("", "glow-edit")
```
meowgorithm, line 220: Most of this I/O type stuff should be done in a command so it doesn't block. The command will return a message which you can then handle in this `update` function. Commands and Messages are are a Bubble Tea thing, see [the command tutorial](https://github.com/charmbracelet/bubbletea/tree/master/tutorials/commands).

ui/pager.go
```diff
@@ -199,6 +202,92 @@ func (m pagerModel) Update(msg tea.Msg) (pagerModel, tea.Cmd) {
...
+
+						// Create a new stash
+						_, err := m.general.cc.StashMarkdown(m.currentDocument.Note, editString)
+
+						if err != nil {
+							cmd := m.showStatusMessage("Couldn't update stash!")
+							cmds = append(cmds, cmd)
+						} else {
+							// Delete the current stash
+							m.general.cc.DeleteMarkdown(m.currentDocument.ID)
```
meowgorithm, line 278: Please handle the return value here.

ui/pager.go
```diff
@@ -199,6 +202,92 @@ func (m pagerModel) Update(msg tea.Msg) (pagerModel, tea.Cmd) {
...
+							cmd := m.showStatusMessage("Updated stash!")
+							cmds = append(cmds, cmd)
+						}
+					} else {
+						err := ioutil.WriteFile(m.currentDocument.localPath, []byte(editString), 0777)
```
meowgorithm, line 283: Ideally we look at the file's original permissions and use the same ones again when we write here.

meowgorithm:
Thanks so much for the PR! I added some comments inline. That's a _super_ good catch about the stash missing update functionality. We're looking at that internally over here.

The command/message stuff can be a bit tricky if you're not familiar with Bubble Tea or Elm, so we're more than happy to lend a hand there. Also don't forget to trim empty lines.

Also note that I'm getting a fatal error (EOF) with [Kakoune](https://kakoune.org), my `EDITOR`, after exiting and attempting to return to vim.

ui/pager.go
shitchell: Ah yeah, I've seen that. Should've realized that would be ideal to use :P

I think both are good--simple status message displayed to the user and a more detailed error written to a log. I had it displaying lots of status messages; for the life of me idk why I chose not to display one for that possible error.

But I actually decided to clean up the code (I think it's a little more cleaned up, anyways) by chopping it up into several different functions. As a result, I think the displaying of status messages is a lot cleaner. If you [prefer this](https://github.com/charmbracelet/glow/compare/master...shitchell:custom) over what I have here, I'll close this PR and open one for the other branch.

ui/pager.go
shitchell: Fair enough. I was under the impression that the permissions were basically ignored if the file already exists, and that parameter only does anything on creating a new file. I'll definitely add that check.

ui/pager.go
shitchell: Thanks! I'll definitely check that out over the weekend.

ui/pager.go
muesli: @shitchell That's actually correct:

WriteFile writes data to a file named by filename. If the file does not exist, WriteFile creates it with permissions perm (before umask); otherwise WriteFile truncates it before writing, without changing permissions.

I still think we should probably specify 0600 as a reasonable default here.

shitchell:
Thanks for the feedback! Yeah I read through a decent bit of bubbletea and charm to help with this, but it doesn't help that I've never really used go before this :P A lot of programming experience, but I'm constantly googling to figure out what everything means or how to do the most trivial thing lol. So super appreciate any pointers or advice; I'm happy to refine this to make sure the code is actually decent quality when it's merged.

I'll have to check that out. I usually use github.com/zyedidia/micro, but that looks like a pretty awesome editor :o Off the top of my head I have no idea why that'd happen. One of the other guys said earlier that he was running into issues launching an editor from within glow as well. ncurses stuff can be pretty delicate, but I'd expect that to cause more display issues than EOF o_O Ah well, that's a Sunday problem :P

Siilwyn:
Just a thought, to simplify the code guessing the editor might not be needed? Since `glow config` already shows an error message when `EDITOR` is not set having the edit functionality do the same isn't that wild.

muesli:
> Just a thought, to simplify the code guessing the editor might not be needed? Since `glow config` already shows an error message when `EDITOR` is not set having the edit functionality do the same isn't that wild.

I tend to agree here. While the editor guessing is a nice touch, I think it opens up a few issues:

1. Which editors do we add to the list?
2. How do we prioritize the list?

In the worst case we launch an editor that the user is entirely unfamiliar with, and doesn't even know how to quit :smile: 

Since we can detect this configuration issue, we should just make the user aware of it.

shitchell:
> * Which editors do we add to the list?
> 
> * How do we prioritize the list?
> In the worst case we launch an editor that the user is entirely unfamiliar with, and doesn't even know how to quit smile

Refusing to launch due to `$EDITOR` being unset is pretty atypical behavior. I can't think of another program that does that. I do agree with the last point. I have no idea why `vim` tends to be the default. The default option when `$EDITOR` is unset, meaning the user is less likely to be a command line aficionado, shouldn't be something that scares them from ever using the command line again :P

I started the list with `micro`, `nano`, and then `nvim`. `micro` and `nvim` aren't installed by default, so if they *are* installed, it's probably because the user purposefully installed them to use them. Also, `micro` and `nano` are super simple and both display keyboard shortcuts at the bottom of the screen by default so the user shouldn't ever get stuck.

*Personally*, I feel like it's a lot friendlier to find / launch an editor. If `$EDITOR` isn't set, it's increasingly likely that the user isn't *super* familiar with the command line, so asking them to configure it might be asking a lot.

A good compromise that some other apps implement is to prompt the user with a simple dialog like:

```
No EDITOR is set. Would you like to:
1) Use nano (a simple editor)
2) Use vi (a complex editor)
3) Search for and use the first available editor
4) Cancel
Type an option (1-4): 
```

But implementing that would be another issue altogether. If others like the dialog idea, I'm happy to start implementing that with this PR and then issue another PR that implements that feature for `glow config` (to minimize coding). Or if it's preferable to make that a separate issue altogether, I can adjust this to conform to the behavior of `glow config` and implement the dialog for both at a later date.

shitchell:
Thinking about it, `glow` would implement that as an ncurses based menu using bubbletea, which is not something I'm prepared to do right now :P So *if* others agree with that option, that's probably best left as a separate issue.

meowgorithm:
For now, let's keep things simple and start with just supporting `EDITOR`.

Bubble Tea is not ncurses-based, by the way.

muesli:
@shitchell Oops? Anything we can help with?

shitchell:
Oh crap. Since I was also working on the live preview enhancement, I decided to do what I should've done in the beginning--move the changes to separate branches. I totally didn't think about how that would break this PR D:

I ended up rewriting it a good bit, anyways, so it should probably still be a totally new PR. I chunked it up into a few separate methods instead of one long case statement :P I haven't been able to spend too much time on it this weekend, though, so I'll probably just create a new PR in a couple days.

My bad! D:

muesli:
@shitchell No worries! Just a heads up: you can still find the old code in your reflog or here: https://github.com/charmbracelet/glow/commit/f7934012502b52cb95247ef3fbc131c31e17ad89

ui/pager.go
shitchell: > Most of this I/O type stuff should be done in a command so it doesn't block. The command will return a message which you can then handle in this `update` function. Commands and Messages are are a Bubble Tea thing, see [the command tutorial](https://github.com/charmbracelet/bubbletea/tree/master/tutorials/commands).

I know this is closed, but this seemed the relevant place to put this. I just got done rewriting this to use tea.Cmd and tea.Msg... and I didn't even think about how making this non-blocking would be bad. Now, as soon as you open the editor, it jumps right back to trying to view the pager. And there's a fight between the editor and glow painting the screen lol.

tl;dr: I'm 98% sure we want this to be blocking.

PR #216, opened 2020-12-07
Fixed showStatusMessage(statusMessage string)

shitchell:
`showStatusMessage(statusMessage string)` should display the string passed on the status line. Instead, "Stashed!" is hardcoded as the only string it will ever display. This fixes the behavior to work as intended.

### DIFF
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -387,7 +387,7 @@ func (m pagerModel) statusBarView(b *strings.Builder) {
 	// Note
 	var note string
 	if showStatusMessage {
-		note = "Stashed!"
+		note = m.statusMessage
 	} else {
 		note = m.currentDocument.Note
 		if len(note) == 0 {


meowgorithm:
Thanks for this! Could you also set the `Stashed!` message in the appropriate place so everything continues to run as normal?

shitchell:
Haha no problem, I changed literally one word :P

Are you talking about stashing a document in the file browsing view? How "Stashed!" and "Delete this item from your stash?" show up in the header? That appears to be behaving as intended since the code explicitly stores those values in the header, but yeah it can be moved to the bottom for consistency's sake. That'll take a little longer. Would you mind accepting this PR since I'm using the status message for both the live update and file editing issues? And you/I can open a new issue for moving all status messages in the file browsing view to the bottom

meowgorithm:
Actually, I assumed `Stashed!` was hardcoded in one place and needed to be moved elsewhere, but it looks like it's was hardcoded in two places. Thanks again for catching this.

PR #217, opened 2020-12-08
Extend example how to read from stdin

nilsvu:
I had to experiment a bit until I found the right way to render from stdin. This extended example could be helpful to someone.


SanJJ1:
can this pr be cleaned up?

maaslalani:
Thanks for this PR @nilsvu, I think this adds some nice context on a real world example of how to use `glow` with stdin!

PR #224, opened 2020-12-16
Expand the path for a specified style file

### LINKED ISSUE #222: Can't use `$HOME` or `~` in config
When setting a style path, you can't reference the user's home directory; you can only use a full path.

*Adding `$HOME` to glow.yml*
```yml
style: "$HOME/.config/glow/styles/shitchell.json"
```

```sh
$ glow test.md
Error: open $HOME/.config/glow/styles/shitchell.json: no such file or directory
```

*Adding `~` to glow.yml*
```yml
style: "~/.config/glow/styles/shitchell.json"
```

```sh
$ glow test.md
Error: open ~/.config/glow/styles/shitchell.json: no such file or directory
```

penguwin:
This expands the specified path for the `style` file in the config. Supporting tilde and values from the current environment variables.

 Now can specify paths like this:
- `~/.config/glow/style.json`
- `$HOME/.config/glow/style.json`
- `${HOME}/.config/glow/style.json`

Resolves #222

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/mattn/go-runewidth v0.0.9
 	github.com/meowgorithm/babyenv v1.3.0
 	github.com/microcosm-cc/bluemonday v1.0.4 // indirect
+	github.com/mitchellh/go-homedir v1.1.0
 	github.com/muesli/gitcha v0.1.2-0.20200908172931-5aa4fdccf2f6
 	github.com/muesli/go-app-paths v0.2.1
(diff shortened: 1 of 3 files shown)


gennaro-tedesco:
Does this work as intended currently? I am trying to specify `$HOME` in any of the above method and it excepts with the same errors reported [here](https://github.com/charmbracelet/glow/issues/222) (which is also the reason why [this](https://github.com/charmbracelet/glow/discussions/662) doesn't work).

Issue #148, opened 2020-05-21
Pager live preview


morpheusthewhite:
A good additional feature would be a pager mode with a live preview, allowing to modify the file and preview its changes in real time (even better would be to have some sort of analogue of the `synctex`, but probably that's out of the scope of this tool)

mschloesser:
I would love to see this feature too! I tried to fiddle with `watch` and some quick and dirty helper scripts but an actual "live preview" would be great.

muesli:
Agreed, that would be nice.

BachoSeven:
This would close #135 as well.

Lisiadito:
I would love to see that feature as well. 
@mschloesser can you share your script until the feature is implemented? Because I didn't get it to work with watch.

shitchell:
I feel like this shouldn't be too hard.

https://medium.com/@skdomino/watch-this-file-watching-in-go-5b5a247cf71f

Using fsnotify, you could create a file watcher to detect any edits to the current document. Any time there's an edit, re-read the file and update `m.currentDocument.Body`. No promises that I won't get distracted or otherwise not get around to it, but I'll *probably* look at this over the weekend.

shitchell:
So we've got 3 parts to this:

1. Initialize a watcher (should only happen once after you launch glow)
2. Add a new filepath to the watcher (should happen each time you view a local file)
3. Remove a filepath from the watcher (should happen each time you stop viewing a local file / return to file list)

@meowgorithm  @muesli

Do you guys have any suggestions where each step should live? I'm *thinking* first add the watcher as an attribute of pagerModel.  `setContent()` gets run every time we both open a new file *and* when we re-render the document (e.g. after detecting a file change). I don't think there presently exists a single method that runs only when we first open a file, so perhaps add another attribute to pager like `isViewingFile bool`, and then adding a block to `setContent()` such as:

```go
if !isViewingFile:
  isViewingFile = true
  if isLocalFile:
    addWatcher(currentFilePath)
```

and then toggle `isViewingFile` to false and remove the watcher whenever you call `unload()`. Does that seem reasonable? Or would there be a better way to implement this? Since I'm still not super familiar with the code, I wanted to ask before implementing it :P

mschloesser:
> I would love to see that feature as well.
> @mschloesser can you share your script until the feature is implemented? Because I didn't get it to work with watch.

@Lisiadito Unfortunately, I deleted the scripts since it didn't really work well. 

[...]

PR #225, opened 2020-12-16
Add option to edit current document

shitchell:
Implements charmbracelet/glow#182

Allows the user to edit the current document by typing 'e'. If `$EDITOR` is set, a temporary file is created with the contents of the current document, and the specified editor is used to open it. If no changes are made when the editor exits, nothing happens. If changes *are* made, then:

**if** the document is a local file, the edits are written to that file
**if** the document is stashed, a new stash is created with the edits, and the old stash is deleted.

It would be ideal to simply update the contents of a stash, but the charm API doesn't currently allow this. Deleting the selected stash has the side effect that, when the user returns to the list of files and tries to re-open the same document, the stash won't load. Restarting `glow` solves this and shows the newly created, updated stash.

https://github.com/charmbracelet/charm/pull/13 creates a `SetMarkdownBody()` method to resolve this minor issue.

### DIFF
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -213,23 +213,6 @@ func (m pagerModel) Update(msg tea.Msg) (pagerModel, tea.Cmd) {
 			}
 		default:
 			switch msg.String() {
-			case "1":
-				m.currentDocument.Body = "Hello world"
-				cmd := m.showStatusMessage("Testing redraw + cmds")
-				cmds = append(cmds, cmd)
-				cmd = renderWithGlamour(m, m.currentDocument.Body)
-				cmds = append(cmds, cmd)
-			case "2":
-				m.currentDocument.Body = "World, hello"
-				cmd := renderWithGlamour(m, m.currentDocument.Body)
-				cmds = append(cmds, cmd)
-				cmd = m.showStatusMessage("Testing redraw + cmds")
-				cmds = append(cmds, cmd)
-			case "r":
-				cmd := m.showStatusMessage("Redrew page!")
-				cmds = append(cmds, cmd)
-				cmd = renderWithGlamour(m, m.currentDocument.Body)
-				cmds = append(cmds, cmd)
 			case "e":
 				editedDocument, err := m.editCurrentDocument()
 
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -656,12 +656,13 @@ func (m pagerModel) editCurrentDocument() (string, string, error) {
 	
 	// Open the temporary file with an editor
 	err = openFileInEditor(tempFile.Name())
-	if err.Error() == "No editor found\n" {
-		return "", "Couldn't find an editor!", err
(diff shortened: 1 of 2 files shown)


shitchell:
So I'm realizing now which parts of this should be (non-)blocking. Launching the editor needs to be blocking. But once the edits are made, saving the local file / stashing the changes needs to be done asynchronously. Working on that now

shitchell:
Alright, as far as I can tell, this is in a pretty good state. Found no bugs while testing. Code is much more organized. Stuff that should block blocks and stuff that shouldn't doesn't.

meowgorithm:
Thank you for the PR!

Just a heads up that we're working on an update to Glow that will most likely conflict with this, though we're happy to look at those conflicts on our end. The holidays are also upon us so it may be a little bit before we give this PR a proper review.

shitchell:
Baha, no worries. I'm happy to look over it, too. If there's a fork or branch where the update is being worked on, I could look over adjusting my code to work with it

meowgorithm:
Cool, we'll push something up early next week. We may contribute to your branch as well!

shitchell:
So I've been testing this out with various editors. It behaves oddly with GUI editors.

* The file remains displayed while the editor is launched. 
  * This might not be a bad thing! If/when charmbracelet/glow#148 is implemented, it might be nice to have a live preview of the document while you're editing it. However, this likely won't work out of the box; we'd probably have to re-work some of the current code to have a live preview while a GUI editor is in use.
* The control isn't necessarily immediately returned to `glow` after you're finished editing. For example, if I use `EDITOR=subl glow` when I'm already using Sublime Text to edit other files... when I'm finished editing the glow document, I don't exit Sublime since I'm still using it for other stuff. Since `glow` is waiting for the editor to exit, I then have to either quit Sublime and interrupt my other work or kill `glow` and lose my edits.

A couple possible solutions:

1. Use `fsnotify` to detect when the temporary file being used for edits is written to. Copy those changes to the local file or upload them to the stashed file as appropriate for every write event.
2. If we can detect whether the editor is GUI based or console based, make GUI file editing non-blocking. Instead, display a message like "File opened in external editor. Press 'enter' to continue when finished.". 

I haven't the slightest idea how to detect if the editor is GUI or console based, but intuition says... while nothing is technically "impossible" when programming, this probably leans towards more complex than it's worth. If that's the case, I currently don't have a good idea for returning control back to `glow` when finished editing.

i3d:
Having a direct editing feature would be awesome!

kinduff:
@shitchell curious if you tried using `EDITOR=subl --wait` in order for the editor to return once the file is saved/closed. Sorry for the necrobump.

maaslalani:
Hey @shitchell, thank you so much for this PR!

Open in editor was added to Glow recently in:
* https://github.com/charmbracelet/glow/pull/361

PR #240, opened 2021-01-31
Removes duplicate doc about stashing

brunoarueira:
The duplicate docs which this pr removes are the same on stashing with CLI, although the section is about the TUI.


muesli:
Thank you!

PR #242, opened 2021-02-03
Enable/Disalbe pager from config

### LINKED ISSUE #231: Set pager option from within glow.yml
It seems we don't have an option to enable pager support inside glow.yml config file.
These are what I've tried:
* pager: true
* pager: ${PAGER}
* pager: "/usr/bin/less"
* pager: "less"

But none of them worked.

lestrrat:
The actual pager command can be passed via PAGER env var

fixes #231

### DIFF
--- a/main.go
+++ b/main.go
@@ -140,6 +140,7 @@ func validateOptions(cmd *cobra.Command) error {
 	width = viper.GetUint("width")
 	localOnly = viper.GetBool("local")
 	mouse = viper.GetBool("mouse")
+	pager = viper.GetBool("pager")
 
 	// validate the glamour style
 	style = viper.GetString("style")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -140,6 +140,7 @@ func validateOptions(cmd *cobra.Command) error {
 	width = viper.GetUint("width")
 	localOnly = viper.GetBool("local")
 	mouse = viper.GetBool("mouse")
+	pager = viper.GetBool("pager")
```
muesli, line 143: I wonder if we should make this a `bool` or let the user specify the pager command as a string here. As it's currently in-line with how the `--pager` flag works, I think this is already an improvement I'll merge, though. Open for discussion nonetheless.

muesli (approved):

PR #253, opened 2021-03-02
Fix some filtering and file listing bugs

### LINKED ISSUE #251: Resetting filter doesn't reset pager
With a filter applied, pressing `ESC` resets the filter. It doesn't however reset the pager's state, limiting the amount of visible items.

meowgorithm:
* Fix pagination when resetting the filter
* Fix inconsistencies when switching between dot and Arabic pager types

Closes #251.

### DIFF
--- a/ui/stash.go
+++ b/ui/stash.go
@@ -64,6 +64,14 @@ const (
 	stashStateShowingError
 )
 
+func (s stashViewState) String() string {
+	return map[stashViewState]string{
+		stashStateReady:           "ready",
+		stashStateLoadingDocument: "loading document",
+		stashStateShowingError:    "showing error",
+	}[s]
+}
+
 // The types of documents we are currently showing to the user.
 type sectionKey int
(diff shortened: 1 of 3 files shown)


ui/stash.go
```diff
@@ -64,6 +64,14 @@ const (
 	stashStateShowingError
 )
 
+func (s stashViewState) String() string {
```
muesli, line 67: Are these `String` functions actually used somewhere yet? I think we may be able to use them to clean up the code a bit.

meowgorithm: I use them in debugging (in fact I used some of them to debug this). I had taken them out at one point, but since I’ve added them back I figured I’d keep them around. Been back and forth about them, though, and they do have the potential for rot, which is an argument to remove them.

meowgorithm: Alright, dropped 'em. Let's keep this code lean.

PR #258, opened 2021-03-09
Pass through all keys when editing filter

muesli:
Otherwise pressing `q` while editing the filter will quit glow.


meowgorithm:
LGTM. Oh glorious <kbd>q</kbd>, it's good to have you back.

PR #270, opened 2021-04-29
Use golang.org/x/term

tklauser:
The golang.org/x/crypto/ssh/terminal package is deprecated and merely a
wrapper around golang.org/x/term. Use the latter directly and avoid a
direct dependency on the former.


muesli:
Thanks again! :heart:

PR #290, opened 2021-08-14
Mention ARM availability for macOS

tisba:
Hi there 👋 

I just noticed that there is an ARM version of glow available for macOS. Maybe this should be mentioned in the README :)


muesli:
Thank you!

PR #298, opened 2021-10-02
Add support for GLOW_CONFIG_PATH environment variable

### LINKED ISSUE #292: Support GLOW_CONFIG environment variable as an alternative to setting --config flag
It will be helpful to have a GLOW_CONFIG environment variable to set the glow config file:

```bash
> GLOW_CONFIG=/path/to/glow.yml glow
```

vinicio:
Hi, everyone!

First of all, thanks for creating and maintaining this incredible tool!

I created this PR to add support for `GLOW_CONFIG` environment variable, like described in the issue linked here.

I had some doubts about updating the README.md file with extra config information, so feel free to give me your opinion.

Thank you!

Close #292

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ it in your favorite $EDITOR. Alternatively you can manually put a file named
 that is, please refer to `glow --help`.
 
 You can also set the location of config file using the `--config` flag or setting
-the `GLOW_CONFIG` environment variable.
+the `GLOW_CONFIG_PATH` environment variable.
 
 Here's an example config:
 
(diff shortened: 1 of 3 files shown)


meowgorithm:
Hi! Just a note that this PR should probably also cover the paths in the `glow config` command when the environment variable is set ([here](https://github.com/charmbracelet/glow/blob/8fc9a300a4494bd6f9dc0255f638c42163a168a4/config_cmd.go)).

Separately, I wonder if the variable should be named `GLOW_CONFIG_PATH`, just so that it's very clear that the variable is pointing to a file, rather than containing configuration itself.

vinicio:
You are right, I forgot about the `config` command. I also isolated the variable name in a single const to make it easier to use.

W1M0R:
Thanks @vinicio for your implementation!

bashbunni:
@muesli if you can give me steps on what to test here I'm happy to help

PR #308, opened 2021-11-22
docs: streamline install instructions

jonaustin:
Add a note to just use `go install` for installation.

Perhaps obvious, but I tend to simply use go as a package manager for golang projects.

Edit: Also, this is just super nice; been looking for a good terminal markdown viewer and this renders beautifully, thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,12 +29,9 @@ More on that below.
 
 ## Installation
 
-Use your fave package manager:
+### Package Manager
 
 ```bash
-# go install
-go install github.com/charmbracelet/glow@latest
-
 # macOS or Linux
 brew install glow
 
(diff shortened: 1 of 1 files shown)


jonaustin:
hmm...I'm fairly new to golang, and after poking at this a bit more; while go install makes installing packages easy; there's apparently no way to easily upgrade those packages after the install without using a 3rd party tool (e.g. https://github.com/TekWizely/bingo )

so I guess I'll close this to avoid confusion when folks want to upgrade after using `go install`.

muesli:
Adding a `go install` command to the README isn't actually the worst idea. Probably below the `go build` instructions as it mostly targets other Go developers & users.

jonaustin:
oh, yeah, I meant to re-open this after I realized that's just normal with Go; thanks, reopened :+1:

jonaustin:
by the way -- all the charm stuff just looks fantastic, great example of why OSS is wonderful!

jonaustin:
@bashbunni Sure thing! I should have a chance to refactor it this weekend.

jonaustin:
@bashbunni Should be good now =)

bashbunni (approved):
Awesome, thanks so much!

PR #343, opened 2022-05-16
Add Chrome OS install option

uberhacker:
See https://github.com/skycocker/chromebrew.


caarlos0:

PR #361, opened 2022-07-15
Open Local Markdown Files in `EDITOR`

maaslalani:
Fixes https://github.com/charmbracelet/glow/issues/182

Introduce key binding to allow users to open their `EDITOR` to edit local
markdown files in stash and pager view.

### DIFF
--- a/ui/editor.go
+++ b/ui/editor.go
@@ -1,0 +1,33 @@
+package ui
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+const defaultEditor = "nano"
+
+type editorFinishedMsg struct{ err error }
+
+func openEditor(path string) tea.Cmd {
+	editor, args := getEditor()
+	cmd := exec.Command(editor, append(args, path)...)
+	cb := func(err error) tea.Msg {
+		return editorFinishedMsg{err}
+	}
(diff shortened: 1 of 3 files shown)


ui/editor.go
```diff
@@ -0,0 +1,22 @@
...
+	"os"
+	"os/exec"
+
+	tea "github.com/charmbracelet/bubbletea"
+)
+
+type editorFinishedMsg struct{ err error }
+
+func openEditor(path string) tea.Cmd {
+	editor := os.Getenv("EDITOR")
```
caarlos0, line 13: needs to handle args too, check #364

ui/editor.go
```diff
@@ -0,0 +1,22 @@
...
+	if editor == "" {
+		editor = "vim"
```
skatkov, line 15: Maybe it's better to change to nano as default? It's very likely that nano is pre-installed on OS, very unlikely that vim will be present.

maaslalani: That's a good point, will change. If someone uses `vim` they should probably have their `$EDITOR` set as well 👍

ui/editor.go
maaslalani: 🫡 Will do, thank you!

caarlos0 (approved):

caarlos0 (approved):

PR #366, opened 2022-08-11
Allow `--width=0` to disable wordwrap

jilleJr:
The wordwrapping makes it much more difficult to copy snippets from the terminal. Especially with code blocks.

The best option would be if I could disable wordwrap only for code blocks, but this is an easy long-hanging fruit.

### DIFF
--- a/main.go
+++ b/main.go
@@ -161,18 +161,20 @@ func validateOptions(cmd *cobra.Command) error {
 	}
 
 	// Detect terminal width
-	if isTerminal && width == 0 && !cmd.Flags().Changed("width") {
-		w, _, err := term.GetSize(int(os.Stdout.Fd()))
-		if err == nil {
-			width = uint(w)
-		}
+	if !cmd.Flags().Changed("width") {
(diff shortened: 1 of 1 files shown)


magnus-ISU:
Markdown shouldn't word wrap in code blocks at all, I think, though CommonMark should be checked.

I've not really used glow much myself because it doesn't seem to care about how markdown should be displayed (hard line breaks also aren't supported despite a PR existing for a month now? with no issues mentioned left to go) but I would suspect that the correct solution here too would be to not require an argument and just print markdown correctly.

I like the suggestion too though.

4l1fe:
Join. If i open with the pager `-p less` i can use it's option `-S`.

    -S or --chop-long-lines
          Causes lines longer than the screen width to be  chopped  (truncated)  rather  than
          wrapped.  That is, the portion of a long line that does not fit in the screen width
          is not shown.  The default is to wrap long lines; that is, display the remainder on
          the next line.

caarlos0:
thanks!

PR #374, opened 2022-09-03
use io or os packages instead of deprecated io/ioutil 

### LINKED ISSUE #373: chore: clean up deprecated io/ioutil dependencies
We are still depending on functionality from the package `io/ioutil`. Run `golangci-lint` on the project to see the where this needs to be changed.

A PR including some tests to ensure no change in functionality would be ideal.

kandros:
fixes #373 

@bashbunni honestly I have no idea what's worth testing in this case, the old packages are being used as a proxy, no specific code used.

```go
// io/ioutils
func ReadAll(r io.Reader) ([]byte, error) {
	return io.ReadAll(r)
}
```

```go
// io/ioutils
func ReadFile(filename string) ([]byte, error) {
	return os.ReadFile(filename)
}
```

Not sure how to handle the go version in go.mod, feedback on that?

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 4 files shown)


muesli:
I'll keep this open until we decide to bump the Go dependency.

caarlos0:
this is fixed on master, thanks for the pr, sorry for the delay

Issue #387, opened 2022-09-29
copy to clipboard


bashbunni:
(Discussed in dev meeting) was asked to add functionality to copy to clipboard
- Use same styling as stashing documents in pager view

caarlos0:
already implemented I believe, closing

PR #388, opened 2022-09-29
feat: copy contents to clipboard

bashbunni:
## Changes:
- [x] copy entire file contents to clipboard with same styling as stashing

The timeout message is time-based like the stash message.

## Related:
#387 
screen capture of UI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/charmbracelet/glow
 go 1.16
 
 require (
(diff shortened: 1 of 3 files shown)


maaslalani:
@bashbunni Tested this out and it works well! Great work!!! I'm going to merge this in https://github.com/charmbracelet/glow/pull/439 because I couldn't get your PR to rebase. Sorry about that.

PR #391, opened 2022-10-09
fix: handle PAGER paths with spaces

### LINKED ISSUE #385: Windows: Doesn't handle $PAGER path with a space
I've installed git for windows and they have less installed here:

C:\Program Files\Git\usr\bin\less.exe

If I add this to the $PAGER eviromental variable, i get the following if I try to use it:
Error: exec: "C:\\Program": file does not exist

Tried to add set a double quote to the string but that didn't seem to help.

glow version 1.4.1 (4863f57)

Shreya-7:
fixes #385 

This fixes the handling of pager paths pulled from environment variables. It works on the assumption that the value of the `PAGER` environment variable will be the path to an executable, and not an actual command (with arguments and what not).

The logic has been moved inside a utility function for the following reasons:
- If, in the future, we need to support more complex logic for building the actual command to be run for using pagers, we can add it in this function instead of bloating our main code.
- Makes the code re-usable, if we need to support such logic for other options (after renaming).
- Makes it easier to test.


Shreya-7:
@bashbunni do you mind reviewing this PR?

Shreya-7:
@muesli I see you added the bug label to this PR. Silly question, but is it because it solves a bug, or are the changes this PR introduced not working as expected? 😐

muesli:
It's there to indicate it addresses a bug in glow.

Shreya-7:
@muesli do the changes look good to you? Do we need something else as well here?

glow_test.go
```diff
@@ -61,3 +65,37 @@ func TestGlowFlags(t *testing.T) {
...
+		{
+			pagerEnvVar:     "",
+			expectedCommand: []string{"less", "-r"},
+		},
+	}
+
+	for _, v := range tt {
+		pager := "PAGER"
+		os.Setenv(pager, v.pagerEnvVar)
+		os.Setenv("PATH", os.Getenv("PATH")+":"+v.pagerEnvVar)
```
muesli, line 95: Do we actually need to update the PATH here? It will get expanded further with each iteration.

glow_test.go
```diff
@@ -61,3 +65,37 @@ func TestGlowFlags(t *testing.T) {
...
+			pagerEnvVar:     "usr/local/bin",
+			expectedCommand: []string{"usr/local/bin"},
+		},
+		{
+			pagerEnvVar:     "",
+			expectedCommand: []string{"less", "-r"},
+		},
+		{
+			pagerEnvVar:     "",
+			expectedCommand: []string{"less", "-r"},
```
muesli, line 88: Unless I'm missing something, the last two cases seem to be dupes.

muesli (changes requested):

muesli:
Looks pretty good already, see the two comments I left tho. Extract credits for adding a test case! :+1:

glow_test.go
Shreya-7: Oop, yes. I had a previous iteration that would take in the command line option as well, and had added it for covering that. Removed, thanks!

glow_test.go
Shreya-7: Not required, had set it for testing a previous iteration, removed.

Shreya-7:
@muesli do you mind giving this a look again?

bashbunni:
Sorry for the delay @Shreya-7 we will get to it, there are just some pending functionality we need to figure out before merging

Shreya-7:
@bashbunni do we still need this bugfix PR?

PR #416, opened 2022-11-12
feature: Refresh local files section using 'r' in pager

deicon:
Using Glow as a dashboard into my local Markdown folder, it would be nice to be able to easily reload the folder content. 

This PR adds a simple refresh using 'r' for local files area only

image

image

image

PR #427, opened 2022-12-07
Feature/poll active file for change

deicon:
Allow automatic refresh from currently rendered and displayed local file. 

New Flags

--watch, -f     Enable File Watching (Defaults false)
--interval, -i  Interval in seconds to poll for file updates

This would fix #426 and is related to #416


gibbz00:
Works like a charm 😍 what's the status on merging this into master?

deicon:
I tried to fix a few bugs and added some enhancements to various charm projects. Unfortunately the response time once a PR is opened is not the best. So as there is no reaction on this PR since Dec 22, I would argue, it will not be merged any time soon

marcelarie:
👀

Wheest:
This is a feature that would be useful to me.

In the meantime, I have achieved a similar (though less effective) solution using the `inotify-tools` package

```sh
MYFILE=log_file.txt
while inotifywait -e modify $MYFILE; do
    clear
    cat $MYFILE | glow
done
```

deicon:
Sure. Or using [entr](https://github.com/eradman/entr). I keep on using my patched version of gum though :)

VanuPhantom:
^ bumping this PR

The fact that this feature has been ready since December last year and still hasn't been added is a shame as it's a wonderful concept!

deicon:
> ^ bumping this PR
> 
> 
> 
> The fact that this feature has been ready since December last year and still hasn't been added is a shame as it's a wonderful concept!

I almost forgot about this open PR. I went from being motivated to work on issues here, to Beeing realistic and did not expected it to be merged soon. Turns out i was right 😀 
I believe I also asked the maintainers if there is anything missing for this to be considered. It might not be on their agenda. But even then I would expect the PR to handled somehow. Rejecting at least.

timskovjacobsen:
This works nicely for me.

I uninstalled the upstream glow and will be using your version for the time being. It's vital to have live rendering while writing a markdown document.

I noticed that the file watcher only works in TUI mode, i.e. when invoked with `glow` and then navigating to the file via the TUI.

I tried to do this at first `glow --watch --interval 1 --pager`. When it did not work reload on changes, I noticed that `glow -mentioned it only works in TUI mode.

deicon:
This may be a bug indeed. Not sure if am willing to spend more time on this to make it not only work in tui mode.

znd4:
> Sure. Or using [entr](https://github.com/eradman/entr). I keep on using my patched version of glow though :)
> 
> `ls *.md | entr glow yourfile.md`

Adding [skim](https://github.com/lotabout/skim) and some extra `entr` functionality for fuzzy-finding and a pretty experience:

```shell
sk | entr -c glow  /_
```

deicon:
@meowgorithm Is there still need for this, or should I close it ?

meowgorithm:
Thanks for flagging this, @deicon. We'll take a look at this and let you know of week.

muesli:
I think a polling solution should only be a fallback for monitoring files. Ideally we use one of the platform's native ways to monitor filesystem events: kqueue, inotify, and so on. There's an `fsnotify` package that abstracts this functionality for each platform, but it doesn't have a polling fallback yet. Maybe this would be a good contribution there?

There are also additional problems that we uncovered while experimenting with filesystem monitoring:

1. Partial writes: we will need to figure out a way to detect when a file has been fully updated, not when writing to it starts. Otherwise you render incomplete content.
2. Some editors & tools will not actually overwrite the contents of a file, but write to a temporary file first, then move it in place. In such situations the original file typically briefly disappears, along with all the monitoring you set up.

timblaktu:
Happy to find @Wheest's solution. I would have never thought to try `cat $FILE | glow` as a workaround for `glow $FILE` not working properly in subshells but working fine in interactive shell. Does anyone know whether this bug is being tracked? Probably some kind of mis-handling of stdin/out or shell environment..

PR #431, opened 2022-12-19
Fix double close body

maaslalani:
This PR fixes an issue with the latest `glow` which double closes the response
body which prevents it from reading HTTPS links.

The `sourceFromArgs` closes the https response body and provides the body to the `source` which also closes it.

```
defer src.reader.Close()
```

This PR removes the initial closing so that the `source` can be read and
closed.


maaslalani:
The response is closed here: https://github.com/charmbracelet/glow/blob/master/main.go#L243

PR #435, opened 2023-01-03
fix: deref url and make combo of readme names

### LINKED ISSUE #434: `glow github.com/owner/repo` doesn't work with lowercase "readme.md" file

jolheiser:
Resolves #434 

As well, fixes a subtle bug where the loop was using a pointer and the resulting URL was incorrect after a single iteration.

### DIFF
--- a/github.go
+++ b/github.go
@@ -29,8 +29,13 @@ func findGitHubREADME(s string) (*source, error) {
 	}
 	u.Host = "raw.githubusercontent.com"
 
+	readmeNamesCombo := make([]string, 0, len(readmeNames)*2)
 	for _, r := range readmeNames {
-		v := u
+		readmeNamesCombo = append(readmeNamesCombo, []string{r, strings.ToLower(r)}...)
+	}
+
+	for _, r := range readmeNamesCombo {
+		v := *u
 		v.Path += "/master/" + r
 
 		// nolint:bodyclose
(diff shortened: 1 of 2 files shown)


danielwerg:
https://github.com/charmbracelet/glow/pull/436 I was late :D

jolheiser:
> #436 I was late :D

Ah sorry! I saw this pop up and was curious, then got nerd-sniped. 😅

danielwerg:
Also while we at it, it would probably be worth supporting `Readme`, `Readme.md`, `ReadMe` and `ReadMe.md`

jolheiser:
> Also while we at it, it would probably be worth supporting `Readme`, `Readme.md`, `ReadMe` and `ReadMe.md`

Due to how this checks naively it might be more work than worth to make a matrix of all possibilities.  
Perhaps an improvement would be to use the API to do a more robust check.

danielwerg:
https://docs.github.com/en/rest/repos/contents#get-a-repository-readme

jolheiser:
> https://docs.github.com/en/rest/repos/contents#get-a-repository-readme

Looks fine to me, could probably even do it without importing the entire `go-github` lib. 
I'm afraid I won't be able to until later, so whether this PR is merged or gets superseded are both fine to me.

danielwerg:
> Looks fine to me, could probably even do it without importing the entire `go-github` lib. I'm afraid I won't be able to until later, so whether this PR is merged or gets superseded are both fine to me.

https://github.com/charmbracelet/glow/pull/438

caarlos0:
this should be fixed by #450

caarlos0:
thanks for the pr, and sorry for the delay!

PR #436, opened 2023-01-03
Fix #434

### LINKED ISSUE #434: `glow github.com/owner/repo` doesn't work with lowercase "readme.md" file

danielwerg:
Fixes https://github.com/charmbracelet/glow/issues/434


danielwerg:
See https://github.com/charmbracelet/glow/pull/435

PR #449, opened 2023-01-24
fix: improve editor handling

### LINKED ISSUE #447: glow does not start without $EDITOR
Hey!

In Glow 1.5 I can now edit a file from my preferred $EDITOR (defaults to `nano`) but [these lines](https://github.com/charmbracelet/glow/blob/4dd3ba1d3c2b47a68b3ee4a7f30f97b0e5b18c78/config_cmd.go#L36-L38) here still require `glow` to read $EDITOR (I don't keep $EDITOR set so I noticed this inconsistency).

Would #361 make that check superflous - perhaps factor out `getEditor()` and use it globally?

thanks for creating `glow`!

caarlos0:
closes #447

### DIFF
--- a/config_cmd.go
+++ b/config_cmd.go
@@ -1,14 +1,12 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
-	"os/exec"
 	"path"
 	"path/filepath"
-	"strings"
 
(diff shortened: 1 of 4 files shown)


maaslalani (approved):
LGTM

PR #450, opened 2023-01-24
fix: check other possible readme paths/branches

### LINKED ISSUE #434: `glow github.com/owner/repo` doesn't work with lowercase "readme.md" file

caarlos0:
closes #434

also check main (as well as master) branches


main.go
```diff
@@ -28,14 +28,15 @@ var (
...
 
-	readmeNames  = []string{"README.md", "README"}
-	configFile   string
-	pager        bool
-	style        string
-	width        uint
-	showAllFiles bool
-	localOnly    bool
-	mouse        bool
+	readmeNames    = []string{"README.md", "README", "readme.md", "readme"}
```
danielwerg, line 31: What about https://gitlab.com/terrakok/gitlab-client?

caarlos0: oh, "Readme.md" is quite common, too, indeed

PR #502, opened 2023-05-23
Added a cmd line switch and config to make WithPreservedNewLines optional

paul-gauthier:
I often use glow to render README.md files that will be shown on GitHub. The GitHub flavor of markdown does not preserve newlines, so I've been wanting a way to set that preference in glow.

Let me know if this is a feature you might want to include, or if I can make any changes to the PR to improve it.


caarlos0:
see https://github.com/charmbracelet/glow/pull/623

PR #505, opened 2023-05-26
Replace string concatenation in loop with string builder

appgurueu:
For large input files, the string concatenation in a hot loop done to trim lines is a bottleneck, running in quadratic time since it has to copy the current string every time. Using a string builder, this can be optimized in linear time.
For Minetest's large [`lua_api.md`](https://github.com/minetest/minetest/blob/master/doc/lua_api.md), this brings the time down from 45s to 3s, a 15x speedup, on my laptop.

### DIFF
--- a/main.go
+++ b/main.go
@@ -284,15 +284,16 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
 
 	// trim lines
 	lines := strings.Split(string(out), "\n")
-	var content string
+	var cb strings.Builder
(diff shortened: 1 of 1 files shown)


LizzyFleckenstein03:
This is indeed a huge speed up.
On my machine rendering minetest's `lua_api.md` takes

- **32.57s** before this change. flamegraph:

flamegraph_before

- **1.94s** after this change. flamegraph:

flamegraph_after

caarlos0 (approved):

PR #527, opened 2023-08-24
fix: handle running inside a snap

caarlos0:
snaps have a myriad of `SNAP_.*` env vars set, so I see if one of them is present (in which case we are sandboxed, and can't really open an EDITOR) and error out to the user suggesting installing it some other way.

### DIFF
--- a/editor/editor.go
+++ b/editor/editor.go
@@ -13,7 +13,7 @@ const defaultEditor = "nano"
 // $EDITOR is set.
 func Cmd(path string) (*exec.Cmd, error) {
 	if os.Getenv("SNAP_REVISION") != "" {
-		return nil, fmt.Errorf("glow is sandboxed, and thus cannot open an editor. To continue using this feature, please install glow using another option")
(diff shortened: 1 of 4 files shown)


bashbunni (approved):
I think this looks good! The error message *might* be helpful to include the fact that Snap is causing the issue, but I think the resolution is the same regardless so not sure it needs to be included.

caarlos0:
yeah messaging definitely can be better, I just can't find a good way to word it... wanna give it a swing @bashbunni ?

editor/editor.go
```diff
@@ -10,9 +11,12 @@ const defaultEditor = "nano"
 
 // Cmd returns a *exec.Cmd editing the given path with $EDITOR or nano if no
 // $EDITOR is set.
-func Cmd(path string) *exec.Cmd {
+func Cmd(path string) (*exec.Cmd, error) {
+	if os.Getenv("SNAP_REVISION") != "" {
+		return nil, fmt.Errorf("Did you install with Snap? Glow is sandboxed and unable to open an editor. Please install Glow with Go or another package manager to enable editing.")
```
bashbunni, line 16: @caarlos0 it's so long, but that's the clearest and most concise I could think of 😂

Issue #535, opened 2023-09-13
Using TUI mode when file name is provided


frederic-martin-doctolib:
Hello,

**Current situation:**

Currently when I open a folder using `glow my-folder` it will list all markdown files in it. This allows me to pick one file and render it. Once the markdown is rendered, I can press 'e' to edit the file using Vim. When I write and quit from vim, I go back to the glow rendering state with the updates. This is very very nice.

**Pain point:**

When I open directly the file using `glow my-folder/myfile.md`, it just does the rendering part but i do not have the edit mode.

Maybe an extra option "-e", like it was done for the pager "-p", could do the trick to enable this new behavior.

WDYT ?

h0rv:
Hi I would like to work on this issue!

andreynering:
Implemented on https://github.com/charmbracelet/glow/pull/679

Issue #547, opened 2023-12-09
Open Editors at Current Line Number in Pager


h0rv:
**Is your feature request related to a problem? Please describe.**

I am currently working on https://github.com/charmbracelet/glow/issues/535, and I think it would be a really neat workflow to 1. Open file with glow with `-e` flag, and edit the file and hop back and forth with the editor quickly and seamlessly.

**Describe the solution you'd like**

I use NeoVim, so for vim-like editors, we can pass in the current line information when opening the editor like so: `vi +25 README.md` to open the file at line 25. 

We would need to check if the editor is Vi/Vim/NeoVim before passing the flag.

We can potentially add support for more editors if they support such a feature.

**Describe alternatives you've considered**

This is potentially not possible if the pager does not have the ability to get the current line number, but I think that is unlikely.

h0rv:
Looks like nano all supports `+<file number>` flag

4l1fe:
Kakoune allows to set the line as well `kak +3192:5 config.yml`

PR #548, opened 2023-12-10
feat: Add `--tui` Flag to Open TUI-mode on File Argument

h0rv:
Adds the feature for the following issue: https://github.com/charmbracelet/glow/issues/535.


h0rv:
Hi @caarlos0, would you be able to review this PR? I will add some tests in the meantime - thanks!

h0rv:
Note: I added optional function parameter (params that have `...<type>`) to existing functions so the existing calls still work. Should I remove that and make them required and just update the other existing calls, or leave it?

ui/ui.go
```diff
@@ -220,13 +227,22 @@ func newModel(cfg Config) tea.Model {
...
 		keygenState: keygenUnstarted,
 		pager:       newPagerModel(&common),
 		stash:       newStashModel(&common),
 	}
+
+	if cfg.FilePath != "" {
+		// Open file passed in with TUI-mode flag
+		m.pager.currentDocument = *localFileToMarkdown(cfg.WorkingDirectory, cfg.FilePath, cfg.FileCreatedAt, true) // TODO: Fix time
+		m.state = stateShowDocument
+		m.pager.update(keyEnter)
```
h0rv, line 242: Hacky solution, but works.

h0rv:
Update `x`, and added functionality to open the editor at the current line as well, resolving https://github.com/charmbracelet/glow/issues/547.

h0rv:
This is the function I want to extract the logic out to open a documentation, so I don't have to simulate the user clicking "enter" on the file they pass in (https://github.com/charmbracelet/glow/pull/548#discussion_r1421678257). However, looking at the `update` function, it would be hard to take the logic out due to the list of commands/messages being collected.

https://github.com/meowgorithm, https://github.com/caarlos0, any thoughts on how to approach this, or is the hack "good enough" for now? Thanks!

h0rv:
Due to rendered spacing, there is some inconsistency to where which line is opened in the editor. Not sure if there is a way to get this exact but it should be good enough as of now.

tarikkirgin:
any update on this?

andreynering:
Hi @h0rv. Sorry for the long wait for a review. This feature was just implemented on https://github.com/charmbracelet/glow/pull/679.

PR #587, opened 2024-03-04
feat: support GLAMOUR_STYLE for custom pager styles

bashbunni:
According to muesli, this was originally a design decision to only support
custom styles with flags, but I think it would be useful to also support custom
themes set with glamour's `GLAMOUR_STYLE` environment variable.


deepanchal:
It would be nice to have this feature. I have to use glow like this currently

```bash
glow -s $GLAMOUR_STYLE
```

since this is not working

```bash
GLAMOUR_STYLE=/tmp/style.json glow
```

Issue #495, opened 2023-05-10
Support `GLOW_CONFIG_DIR` env var


cristiand391:
hi 👋🏼 

I'm trying to consolidate my dotfiles to be the same on macos and linux and have a small issue with how glow handles its config file.

I have a script that runs `yq --inplace '.style = "dark"' ~/.config/glow/glow.yml` when switching between dark/light in my OS to set the right theme for glow but can't reuse the same script on mac and linux  because on mac the path is `~/Library/Preferences/glow/glow.yml` vs `~/.config/glow/glow.yml` on linux.
[I saw this issue requesting to support XDG paths](https://github.com/charmbracelet/glow/issues/328) and then found that glow uses `go-app-paths` to handle platform-specific paths so this should be fixed/added there to work on mac (XDG is supported on linux).

This feature request is to allow users to specify a path for the config file via an env var like `GLOW_CONFIG_DIR`.
It wouldn't be a breaking change like adding support for XDG  on macos and seems easy to add (happy to open a PR  if you'll agree on this request :) )

Subjective:
I second this, as it stands currently glow's config is the only remaining config I have in the ~/Library/Preferences folder. It be great if we could have the option to customize the path  :)

ptrck0:
Made a [PR](https://github.com/charmbracelet/glow/pull/594) to implement this.

PR #594, opened 2024-03-16
Add an environment variable to control the config path.

ptrck0:
Not everyone wants to use Mac OS's default config path "~/Library/Preferences/" as returned by [go-paths](https://github.com/muesli/go-app-paths?tab=readme-ov-file#directories).

This patch includes the environment variable ```GLOW_CONFIG_HOME``` to set, so Mac OS users don't have to run ```glow --config ~/.config/glow/glow.yml``` or something like that.
If you set the config variable, it is used, if not, the original flow is used.

Oh, and a small spelling error fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,8 +193,12 @@ to find more styles. Or [make your own](https://github.com/charmbracelet/glamour
 If you find yourself supplying the same flags to `glow` all the time, it's
 probably a good idea to create a config file. Run `glow config`, which will open
 it in your favorite $EDITOR. Alternatively you can manually put a file named
-`glow.yml` in the default config path of you platform. If you're not sure where
+`glow.yml` in the default config path of your platform. If you're not sure where
 that is, please refer to `glow --help`.
+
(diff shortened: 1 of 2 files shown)


ptrck0:
This will close [this](https://github.com/charmbracelet/glow/issues/495) issue.

jesseb34r:
I would love to see this approved. I consolidate my dotfiles on macos all to `~/.config`.

Additionally it might be worth respecting the `XDG_CONFIG_HOME` environment variable by default which would help glow use the user's preferred config directory more accurately without requiring `GLOW_CONFIG_HOME` to be set.

ccoVeille:
I didn't check at glow code for now, but I expect it to use

https://pkg.go.dev/os#UserConfigDir

https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/os/file.go;l=509

So you would like to overload it when the XDG_CONFIG_HOME env variable is present?

PR #616, opened 2024-06-12
Add flox install

bryanhonof:
`glow` seems to be available and working in Flox today.
If you happen to have any questions about Flox, please, do ask. 😊


caarlos0:

PR #621, opened 2024-07-03
feat: allow config from CHARM_CONFIG_HOME and XDG_CONFIG_HOME

### LINKED ISSUE #328: Support XDG paths on macOS
It'd be great if glow would read its config from `XDG_CONFIG_HOME` instead of `~/Library/Preferences/glow/glow.yml` on macOS.

I frankly don't know anyone who uses `~/Library` for any kind of configuration for CLI tools.

### LINKED ISSUE #475: Respect XDG Base Directory specification on macOS
I was trying out Glow because of a user-reported issue on Ranger earlier and noticed that Glow defaults to writing configuration to `~/Library` on macOS.

While the Freedesktop [XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) isn't really targeted at macOS, so many programs do respect the specification and make it a lot more convenient to manage configurations and such.

Could it be considered that Glow use the XDG Base Directories if the variables are defined in the environment and fall back to the usual location in `~/Library` otherwise?

Another minor note is that I was surprised Glow unconditionally writes a config file when running the `glow config` subcommand, even if no settings are changed. In general I prefer to configure only the things I specifically care about and fall back to defaults, which may change over time, determined to be best by upstream. Maybe the subcommand could check if any changes were made and only write the config if that was the case? And maybe it should offer all the possible settings commented out so you can set just the ones you care about and fall back to defaults for everything else?

### LINKED ISSUE #495: Support `GLOW_CONFIG_DIR` env var

caarlos0:
closes #594
closes #495
closes #328
closes #475

### DIFF
--- a/config_cmd.go
+++ b/config_cmd.go
@@ -7,8 +7,8 @@ import (
 	"path/filepath"
 
 	"github.com/charmbracelet/x/editor"
-	gap "github.com/muesli/go-app-paths"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 const defaultConfig = `# style name or JSON path (default "auto")
@@ -20,18 +20,6 @@ pager: false
 # word-wrap at width
 width: 80
 `
-
-func defaultConfigFile() string {
-	scope := gap.NewScope(gap.User, "glow")
-	if c := os.Getenv("CHARM_CONFIG_HOME"); c != "" {
-		return filepath.Join(c, "glow.yml")
-	}
-	if c := os.Getenv("XDG_CONFIG_HOME"); c != "" {
-		return filepath.Join(c, "glow.yml")
-	}
-	path, _ := scope.ConfigPath("glow.yml")
-	return path
-}
 
 var configCmd = &cobra.Command{
 	Use:     "config",
@@ -63,7 +51,7 @@ var configCmd = &cobra.Command{
 
 func ensureConfigFile() error {
 	if configFile == "" {
-		configFile = defaultConfigFile()
+		configFile = viper.GetViper().ConfigFileUsed()
 		if err := os.MkdirAll(filepath.Dir(configFile), 0o755); err != nil {
 			return fmt.Errorf("Could not write config file: %w", err)
 		}
--- a/config_cmd.go
+++ b/config_cmd.go
@@ -23,6 +23,12 @@ width: 80
 
 func defaultConfigFile() string {
 	scope := gap.NewScope(gap.User, "glow")
+	if c := os.Getenv("CHARM_CONFIG_HOME"); c != "" {
+		return filepath.Join(c, "glow.yml")
+	}
+	if c := os.Getenv("XDG_CONFIG_HOME"); c != "" {
+		return filepath.Join(c, "glow.yml")
+	}
 	path, _ := scope.ConfigPath("glow.yml")
 	return path
 }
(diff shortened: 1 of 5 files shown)


config_cmd.go
```diff
@@ -23,6 +23,12 @@ width: 80
 
 func defaultConfigFile() string {
 	scope := gap.NewScope(gap.User, "glow")
+	if c := os.Getenv("CHARM_CONFIG_HOME"); c != "" {
+		return filepath.Join(c, "glow.yml")
+	}
+	if c := os.Getenv("XDG_CONFIG_HOME"); c != "" {
```
ccoVeille, line 29: Why not using os.UseConfigDir ?

https://pkg.go.dev/os#UserConfigDir

It's implementation might be what you are looking for

https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/os/file.go;l=509

Delta456: UserConfigDir also looks at the home dir `os.GetEnv("HOME")`  if `os.GetEnv("XDG_CONFIG_HOME")` is empty which isn't wanted here.

ccoVeille: Indeed

caarlos0: > On Darwin, it returns $HOME/Library/Application Support

I think it wouldn't change much in that particular problem.

We can work on replacing the usage of gap with this library though

ccoVeille: Please note, I know some macOS users expect to use this folder by default, but want code to respect XDG_CONFIG_HOME if present.

- https://github.com/charmbracelet/glow/pull/594#issuecomment-2189627733

caarlos0: yes, this is what this PR does...

main.go
```diff
@@ -388,13 +388,22 @@ func init() {
 }
 
 func tryLoadConfigFromDefaultPlaces() {
+	fmt.Println("AQUI 1")
```
ccoVeille, line 391: This one was forgotten

main.go
```diff
@@ -388,13 +388,22 @@ func init() {
...
 		fmt.Println("Could not load find config directory.")
 		os.Exit(1)
 	}
 
+	if c := os.Getenv("CHARM_CONFIG_HOME"); c != "" {
+		viper.AddConfigPath(c)
+	}
+
+	if c := os.Getenv("XDG_CONFIG_HOME"); c != "" {
+		viper.AddConfigPath(filepath.Join(c, "glow"))
```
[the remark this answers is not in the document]
caarlos0: it should respect order yet

PR #626, opened 2024-07-09
fix: pass color profile down to glamour

caarlos0:
glamour was ignoring the detected color profile when running `glow file.md`

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/charmbracelet/glamour"
 	"github.com/charmbracelet/glow/ui"
 	"github.com/charmbracelet/glow/utils"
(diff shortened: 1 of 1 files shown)


folliehiyuki:
Sorry to bother you, @caarlos0 ! This commit breaks the workflow of using `glow` as Markdown previewer inside a terminal file manager. `termenv` determines the preview pane being a non-TTY output (which makes sense though).

Can `glow` allow the `-s` flag to bypass ColorProfile, so it can render color codes regardless of stdout?

Here is a screenshot using [vifm](https://github.com/vifm/vifm). The same also happens when using [ranger](https://github.com/ranger/ranger):

PR #631, opened 2024-07-18
feat: updating glamour

caarlos0:
wip while I test the changes of https://github.com/charmbracelet/glamour/pull/317


main.go
```diff
@@ -287,18 +287,6 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
 		return err
 	}
 
-	// trim lines
-	lines := strings.Split(out, "\n")
```
caarlos0, line 291: not sure why this was needed, but it was breaking some things... I removed it and see no difference otherwise 🤔

PR #648, opened 2024-09-13
Enable "r" in filter state

### LINKED ISSUE #644: Can't input `r` when filtering
This is because <kbd>r</kbd> is triggering the (new-ish) reload functionality.

NijeboerFrank:
Fixes #644 

First of all, thank you for this tool! Great fun to use. I was a bit annoyed by it triggering a reload when I'm inputting a filter. To fix it I just copied the way in which this was done from the `q` key. If you want it done differently or have another fix in the works, let me know!


andreynering (approved):
Looks good, thank you @NijeboerFrank!

PR #651, opened 2024-09-29
Make --all configurable globally

### LINKED ISSUE #650: It's surprising that markdown files that are .gitignor'ed are not shown by default (and slightly annoying)
**Is your feature request related to a problem? Please describe.**
Really love 2.0 release! Plenty of cool improvements packaged, some features I never used removed -- it's awesome.

But there is one thing, that caused a lot of confusion and took me at least an hour of nervously looking through source code to figure out what happened.

My confusion has been caused by these lines for code
https://github.com/charmbracelet/glow/blob/master/ui/ui.go#L345-L349

The fact, that .gitignored markdown files are not shown anymore **by default**. This was not the case before, because I actively used previous version of glow to review files that have been .gitignored.

I might have missed a note/change about it and this might be my fault.

**Describe the solution you'd like**
It's particularly annoying, that now I have to pass --all/-a flag to do what I did before without that. I don't feel that this is a good **default** behavior.

But it would be nice if there was an option to permanently change this behavior through glow.yml

As example with following setting:
```yaml
all: true
```

**Describe alternatives you've considered**
Switch default behavior to show all files.

skatkov:
fixes #650

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ mouse: true
 pager: true
 # at which column should we word wrap?
 width: 80
+# show all files, including hidden and ignored.
+all: true
 ```
 
 ## Feedback
--- a/config_cmd.go
+++ b/config_cmd.go
@@ -21,6 +21,8 @@ mouse: false
 pager: false
 # word-wrap at width
 width: 80
+# show all files, including hidden and ignored.
+all: true
 `
 
 var configCmd = &cobra.Command{
--- a/main.go
+++ b/main.go
@@ -36,6 +36,7 @@ var (
 	pager            bool
 	style            string
 	width            uint
+	all              bool
 	showAllFiles     bool
 	showLineNumbers  bool
 	preserveNewLines bool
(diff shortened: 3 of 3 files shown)


bashbunni:
Hey, I saw your issue and this change totally makes sense. We should probably mention this feature in the [config section](https://github.com/charmbracelet/glow/tree/master?tab=readme-ov-file#the-config-file) of the `README`. We should also add this field to the default config as well. Probably defaulting to all? I think that not being able to view files you would expect is more confusing than seeing these files and figuring out how to hide them

skatkov:
@bashbunni that's awesome suggestion!

bashbunni:
I just tweaked some things here - the `all` field was redundant as the `showAllFiles` field was already handling this behaviour. I added a default value of `true` to both the configuration file, and to the field itself in case it's unset anywhere. Tested that it works with both `true` and `false` values both in the config file and the args :)

bashbunni:
Let me know if it behaves as you'd expect, if so I can merge it :)

skatkov:
@bashbunni tested this locally, and it does work as expected!

PR #653, opened 2024-10-16
fix: allow config file to load `line-numbers` and `preserve-new-lines`

prithvijj:
## Changes

- Added in a fix to allow the `line-numbers` and `preserve-new-lines` to be loaded from the config file
- Previously it seems unclear on how to setup the `line-numbers` key within the config, and while there was support added for configuring the `line-numbers` flag, it couldn't be read from the config file, since the `validateOptions` method would not retrieve the key from the config file
- Hence changed it so that it does retrieve the `line-numbers` key from the config file, and use it as needed
- There is a similar issue in `preserveNewLines` option as well, and hence fixed that too
- Updated the README.md to also provide an example on how to setup both the keys

## Other Notes

Related to

- [ ] Any tests that I can write?


prithvijj:
Hey @meowgorithm, sorry for the super late rebase, but should be ready again!
I did a quick sanity test on my local machine with my above testing notes, and does pass,
Feel free to try it out and review it whenever

README.md
```diff
@@ -209,6 +209,10 @@ pager: true
 width: 80
 # show all files, including hidden and ignored.
 all: false
+# show line numbers (TUI-mode only)
+showLineNumbers: false
+# preserve newlines in the output
+preserveNewLines: false
```
andreynering, line 215: I just wonder if it should be `showLineNumbers ` or `show_line_numbers `.

`preserveNewLines` is already camel case, though, so changing that would be a breaking change.

Other than this small decision, looks good to me.

andreynering (approved):

prithvijj:
@andreynering I was wondering what's the merge process now that it's been approved?

andreynering:
@prithvijj Merged. Thank you for contributing!

prithvijj:
Sweet, thank you for reviewing and merging this! 
Appreciate it :)

Shane-XB-Qian:
sry, but this linenum maybe not much useful, seems the linenum maybe not accurate or reflect to real linenum,
or different like the one in 'less', there some blank line or render (of markdown) may render empty/blank lines.

-- 
shane.xb.qian

Issue #660, opened 2024-11-29
Make build work offline


ottok:
While packaging Glow for Debian, I noticed that the build fails as it is not designed to be able to run offline.

The Debian builds run in a hermetic environment that has no external
access neither in nor out. This ensures high level of reproducibility,
stability and security. Glow should refactor these tests so that they
only run if the build is called with some special debug/test tag, or
have a pre-test that skips the tests automatically if there is no network
connection at all.

Thanks!

bashbunni:
Hey, thanks for bringing this to our attention. I've just created a PR to address this

PR #661, opened 2024-12-03
test: skip networked tests if no connection

bashbunni:
after all that I just deleted the networked test 😂

### DIFF
--- a/glow_test.go
+++ b/glow_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"strings"
 	"testing"
 )
 
@@ -16,6 +17,13 @@ func TestGlowSources(t *testing.T) {
 
 	for _, v := range tt {
 		t.Run(v, func(t *testing.T) {
+			// Start by checking for network issues.
+			_, nerr := readmeURL(v)
+			if nerr != nil && strings.Contains(nerr.Error(), "no such host") {
+				t.Logf("Error during execution (args: %s):\n%v", v, nerr)
+				t.Skip("Test uses network. Are you connected to the Internet?")
+			}
+
 			buf := &bytes.Buffer{}
 			err := executeArg(rootCmd, v, buf)
 			if err != nil {
--- a/glow_test.go
+++ b/glow_test.go
@@ -1,40 +1,8 @@
 package main
 
 import (
-	"bytes"
-	"errors"
-	"net"
 	"testing"
 )
-
-func TestGlowSources(t *testing.T) {
-	tt := []string{
-		".",
-		"README.md",
-		"github.com/charmbracelet/glow",
-		"github://charmbracelet/glow",
-		"https://github.com/charmbracelet/glow",
-	}
-
-	for _, v := range tt {
-		t.Run(v, func(t *testing.T) {
-			buf := &bytes.Buffer{}
-			err := executeArg(rootCmd, v, buf)
-			if err != nil {
-				// Check for network issues.
-				var netErr *net.DNSError
-				if errors.As(err, &netErr) {
-					t.Logf("Error during execution (args: %s): %v", v, err)
-					t.Skip("Test uses network. Are you connected to the Internet?")
-				}
-				t.Errorf("Error during execution (args: %s): %v", v, err)
-			}
-			if buf.Len() == 0 {
-				t.Errorf("Output buffer should not be empty (args: %s)", v)
-			}
-		})
-	}
-}
 
 func TestGlowFlags(t *testing.T) {
 	tt := []struct {
--- a/glow_test.go
+++ b/glow_test.go
@@ -2,7 +2,8 @@ package main
 
 import (
 	"bytes"
-	"strings"
+	"errors"
+	"net"
 	"testing"
 )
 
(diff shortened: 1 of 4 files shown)


bashbunni:
Hey @caarlos0 or @aymanbagabas I may need some help getting CI fixed here. It seems to be detecting `glow/v2` as the working directory but I'm not sure how to fix that 🤔

ottok:
Thanks for looking into this. If you append "(Closes: #660)" in the git commit title or put "Closes: #660" anywhere in the git commit message body, it will automatically link to that issue, and also close that issue if the commit gets merged on branch "master".

caarlos0:
> Hey @caarlos0 or @aymanbagabas I may need some help getting CI fixed here. It seems to be detecting `github.com/charmbracelet/glow/v2` as the working directory but I'm not sure how to fix that 🤔

I looked into this, but not sure why its failing..

FWIW we have the url tests that are also skipped due to weird network stuff... maybe worth just removing them

glow_test.go
```diff
@@ -16,6 +17,13 @@ func TestGlowSources(t *testing.T) {
 
 	for _, v := range tt {
 		t.Run(v, func(t *testing.T) {
+			// Start by checking for network issues.
+			_, nerr := readmeURL(v)
+			if nerr != nil && strings.Contains(nerr.Error(), "no such host") {
+				t.Logf("Error during execution (args: %s):\n%v", v, nerr)
+				t.Skip("Test uses network. Are you connected to the Internet?")
+			}
```
ccoVeille, line 25: Please try this, that would be safer

```suggestion
			// Start by checking for network issues.
			_, err := readmeURL(v)
			var dnsError *net.DNSError
			if errors.As(err, &dnsError) {
				t.Logf("Error during execution (args: %s):\n%v", v, err)
				t.Skip("Test uses network. Are you connected to the Internet?")
			}
```

This would be a more idiomatic way.

ccoVeille: I didn't test it, I simply wrote it after checking http.Get stack

I'm unsure if it has to be `if errors.As(err, &dnsError) {` or `if errors.As(err, dnsError) {` as dnsError is a pointer.

I let you check and adapt

bashbunni: Thank you!! This is very helpful :)

glow_test.go
```diff
@@ -16,13 +18,21 @@ func TestGlowSources(t *testing.T) {
...
+				if errors.As(err, &urlErr) {
+					t.Logf("Error during execution (args: %s):\n%v", v, err)
+					t.Skip("Test uses network. Are you connected to the Internet?")
+				} else {
+					t.Errorf("Error during execution (args: %s): %T %v", v, err, err)
+				}
+
+				if buf.Len() == 0 {
+					t.Errorf("Output buffer should not be empty (args: %s)", v)
+				}
```
ccoVeille, line 35: This could be easier to read

```suggestion
				// Check for network issues.
				if errors.As(err, &urlErr) {
					t.Logf("Error during execution (args: %s):\n%v", v, err)
					t.Skip("Test uses network. Are you connected to the Internet?")
					return
				}
				t.Errorf("Error during execution (args: %s): %T %v", v, err, err)

				if buf.Len() == 0 {
					t.Errorf("Output buffer should not be empty (args: %s)", v)
				}
```

glow_test.go
```diff
@@ -16,13 +18,21 @@ func TestGlowSources(t *testing.T) {
 
 	for _, v := range tt {
 		t.Run(v, func(t *testing.T) {
+			var urlErr *net.DNSError
```
ccoVeille, line 21: please move the variable definition aside the errors.As

main.go
```diff
@@ -75,10 +76,14 @@ func sourceFromArg(arg string) (*source, error) {
...
 
 	// a GitHub or GitLab URL (even without the protocol):
+	var dnsErr *net.DNSError
 	src, err := readmeURL(arg)
 	if src != nil && err == nil {
 		// if there's an error, try next methods...
 		return src, nil
+	} else if err != nil && errors.As(err, &dnsErr) {
+		// fail if there's a network error.
+		return src, err
```
ccoVeille: The code could be clearer

```suggestion
	// a GitHub or GitLab URL (even without the protocol):
	src, err := readmeURL(arg)
	if src != nil && err == nil {
		// if there's an error, try next methods...
		return src, nil
	}
	
	var dnsErr *net.DNSError
	if errors.As(err, &dnsErr) {
		// fail if there's a network error.
		return src, err
```

- The `else` is not needed as you use `return` in the `if`
- no need to check `err != nil && errors.As(err`, see https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/errors/wrap.go;l=97-100

main.go
```diff
@@ -75,10 +76,14 @@ func sourceFromArg(arg string) (*source, error) {
 	}
 
 	// a GitHub or GitLab URL (even without the protocol):
+	var dnsErr *net.DNSError
 	src, err := readmeURL(arg)
 	if src != nil && err == nil {
 		// if there's an error, try next methods...
```
ccoVeille, line 86: I'm doubtful about this comment

there is no error, so why talking about trying the other one ?

It worked, no ?

ccoVeille (changes requested):

main.go
```diff
@@ -77,11 +77,12 @@ func sourceFromArg(arg string) (*source, error) {
 
 	// a GitHub or GitLab URL (even without the protocol):
 	var dnsErr *net.DNSError
+	var opErr *net.OpError
```
ccoVeille, line 80: if you are testing net.OpError

then you shouldn't test net.DNSError too

Here are some explanations:
https://stackoverflow.com/questions/69568234/how-do-i-work-out-what-type-of-error-my-error-is-in-go
https://haisum.github.io/2021/08/14/2021-golang-http-errors/

maybe url.Error could be enough for you

glow_test.go
bashbunni: Ah, I'm just debugging github actions right now, will be undoing these commits once I figure out what's going on 😅

bashbunni:
ci is failing because it's getting a forbidden error from the `http.Get` request

PR #667, opened 2024-12-21
fix: adjust validateOptions to handle preconfigured width

### LINKED ISSUE #666: width calculation logic has the tacit assumption that no confiig file exists
**Describe the bug**

When glow is called by the user for the very first time (so that width is not given explicitly via `-w`), glow output with will be inferred from the terminal width (as I suppose it's intended). However, subsequent repetitions of the invocation will have glow output width at 80.

**Setup**

glow between v2.0.0 and  7b3bfac (master as of writing this).

**To Reproduce**

Open a shell in a terminal so that its width is way below 80.

See the line widtths when glow output witdth is expilicty set to that of the terminal:

```sh
$ unbuffer glow -s notty -w $(stty size | awk '{print $2}') sample.md | \
  ruby -ne 'puts $_.size' | sort | uniq
1
28
29
31
```
(Note: the width does not stricty adhere to the specified value  - I think I'll submit another PR for thart - but it's at least close to the speicified witdh.)

Expected behavior without using `-w`:

```sh
$ unbuffer glow -s notty sample.md | ruby -ne 'puts $_.size' | sort | uniq
1
28
29
31
```

Actual behavior:

```sh
$ unbuffer glow -s notty sample.md | ruby -ne 'puts $_.size' | sort | uniq
1
88
89
```

Except for the case when the user runs this command as their _very first_ invocation of glow - in that case we do get the expected output.

**Additional context**

The difference between very first invocation and subsequent ones is that upon that very first invocation glow deploys a basic configuration file. And the code that calculates the actual runtime witdth in `validateOptions` behaves as if it were granted that no config file exists (which holds only for the very first invocation):

https://github.com/charmbracelet/glow/blob/7b3bfac3068c5501adc83c2c1a3914c84b0cb136/main.go#L180-L192

`width == 0` is used in the sense that `width` is "pristine". However, after the deployment of the config file, which includes a setting of `width` to 80, if `-w` is not passed, the execution flow will arrive to this code block with `width` already being 80. Ie. the pristine value is whatever what's been placed into config - a static value check cannot capture this condition.

csabahenk:
fixes #666

### DIFF
--- a/main.go
+++ b/main.go
@@ -175,22 +175,26 @@ func validateOptions(cmd *cobra.Command) error {
 		style = "notty"
 	}
 
-	// Detect terminal width
-	if !cmd.Flags().Changed("width") {
-		if isTerminal && width == 0 {
-			w, _, err := term.GetSize(int(os.Stdout.Fd()))
-			if err == nil {
-				width = uint(w)
-			}
+	// Determining width
 
-			if width > 120 {
-				width = 120
-			}
-		}
-		if width == 0 {
-			width = 80
+	// accept the value of width if it comes from commandline
+	if cmd.Flags().Changed("width") {
+		return nil
+	}
+	// trying to set width to terminal width
+	if isTerminal {
+		w, _, err := term.GetSize(int(os.Stdout.Fd()))
+		if err == nil {
+			width = uint(min(w, 120))
+			return nil
 		}
 	}
+	// accept the value of width if it comes from config
+	if viper.InConfig("width") {
+		return nil
+	}
+	// fall back to a hard-coded sane value
+	width = 80
 	return nil
 }
 


main.go
```diff
@@ -156,6 +156,7 @@ func validateStyle(style string) error {
 
 func validateOptions(cmd *cobra.Command) error {
 	// grab config values from Viper
+	width_in_config := viper.InConfig("width")
```
ccoVeille, line 159:
```suggestion
	widthInConfig := viper.InConfig("width")
```

main.go
```diff
@@ -156,6 +156,7 @@ func validateStyle(style string) error {
 
 func validateOptions(cmd *cobra.Command) error {
```
ccoVeille, line 157: Typo in commit and PR title

validateOptions not validaOptions

main.go
```diff
@@ -177,18 +178,24 @@ func validateOptions(cmd *cobra.Command) error {
...
-			width = 80
+		if w == -1 {
+			if !width_in_config {
+				width = 80
+			}
+		} else {
+			width = uint(w)
 		}
 	}
 	return nil
```
ccoVeille, line 201: What about this:

```go
func validateOptions(cmd *cobra.Command) error {
	// grab config values from Viper
	widthInConfig := viper.InConfig("width")
	width = viper.GetUint("width")
	mouse = viper.GetBool("mouse")
	pager = viper.GetBool("pager")
	showAllFiles = viper.GetBool("all")
	preserveNewLines = viper.GetBool("preserveNewLines")

	// validate the glamour style
	style = viper.GetString("style")
	if err := validateStyle(style); err != nil {
		return err
	}

	isTerminal := term.IsTerminal(int(os.Stdout.Fd()))
	// We want to use a special no-TTY style, when stdout is not a terminal
	// and there was no specific style passed by arg
	if !isTerminal && !cmd.Flags().Changed("style") {
		style = "notty"
	}

	// Detect terminal width
	if cmd.Flags().Changed("width") {
		return nil
	}
	if isTerminal {
		w, _, err := term.GetSize(int(os.Stdout.Fd()))
		if err == nil {
			width = uint(max(w, 120))
			return nil
		}
	}
	if !widthInConfig {
		width = 80
	}
	return nil
}
```

I wrote it on my phone in text mode, please test it and adapt

ccoVeille (changes requested):

main.go
csabahenk: Fine and dandy, apart from having to have `min(w, 120)`, not `max` :)

ccoVeille: I'm glad you find it dandy 😁

Adding a Co-authored-by tag to your commit would be appreciated then

```
Co-authored-by: ccoVeille <3875889+ccoVeille@users.noreply.github.com>
```

Indeed for the min vs max

Issue #669, opened 2025-01-01
respect .gitignore


laoshaw:
**Is your feature request related to a problem? Please describe.**

glow -a will list all markdown files under .venv etc

**Describe the solution you'd like**

make it honor .gitignore

**Describe alternatives you've considered**

**Additional context**

PR #677, opened 2025-02-03
fix: make `all` setting `false` by default

andreynering:
Closes https://github.com/charmbracelet/glow/issues/669


raphamorim (approved):

caarlos0:
yeah, I agree that false should probably be the default, as it will follow gitignore by default.


users that want to see all files can still do so by changing settings or `glow --all` (which I think we might need to better document).


all in all im fine with this :)

bashbunni:
Since this is a breaking change for users, should we consider this for a next major version or some v0 release? I know we did v1 for glow already

andreynering:
@bashbunni Regarding being a breaking-change, it seems to be only partially breaking to me, because for users that already had the configuration file generated on their machine (on the first run), that setting won't be changed automatically. It'll only have the new default on new machines.

No strong opinion, though. I think I vote to merge now, but it's fine to wait if needed.

aymanbagabas (approved):

bashbunni:
@andreynering ah that makes sense, sounds good to me then :)

PR #678, opened 2025-02-03
fix: restore hint for `e` / `edit` on the footer

andreynering:
I found this while working on an unrelated issue. There was not hint for `e`/ `edit` at all.

### Before

Screenshot 2025-02-03 at 17 56 47

### After

Screenshot 2025-02-03 at 17 57 00

### DIFF
--- a/ui/stashhelp.go
+++ b/ui/stashhelp.go
@@ -100,7 +100,6 @@ func (m stashModel) helpView() (string, int) {
 	}
 
 	var (
-		isEditable    bool
 		navHelp       []string
 		filterHelp    []string
 		selectionHelp []string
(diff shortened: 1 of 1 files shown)


ui/stashhelp.go
```diff
@@ -100,7 +100,6 @@ func (m stashModel) helpView() (string, int) {
 	}
 
 	var (
-		isEditable    bool
```
andreynering, line 103: In the code as it is in `master`, this variable was simply never set. It is a left-over from previous changes.

In the past, it was set [like this](https://github.com/charmbracelet/glow/pull/442/files#diff-8d8107f924235e9ef0d2002f9b3088317067bccace395463b8bb62c3dcf8cf2cR126);

```go
isEditable = md != nil && md.docType == LocalDoc && md.localPath != ""
```

That's not easily addable to our current code because it was refactored since.

Question: In which situation a file would not be editable? Is it fine to always show the hint?

raphamorim: cc @caarlos0 @aymanbagabas

raphamorim (commented):
overall looks fine just wonder about edit itself, if was supposed to be toggled. added @caarlos0 @aymanbagabas to confirm

raphamorim (approved):

bashbunni:
Yeah this looks good, just tested locally as well

bashbunni (approved):

aymanbagabas (commented):
LGTM

Issue #489, opened 2023-04-17
How to start glow with a file open (TUI mode)


Disonantemus:
**Glow** has a nice _glamour style_, but only starts in **TUI mode** launching without any markdown files. I have my own _shell script_ with **fd** and **fzf** to choose a file and send it to glow, but only shows CLI output, then it would be more useful have an option (maybe "-t") to start already in **TUI mode**.  

I use **nnn** file manager and  **glow** should work, because it's already in  _nuke_ (plugin), but only get _Escape codes_ like this:

    ESC[38;5;228;48;5;63;1mESC[0mESC[38;5;228;48;5;63;1mESC[0m  ESC[38;5;228;48;5;63;1m ESC[0mESC[38;5;228;48;5;63;1
    
I don't know if I'm doing something wrong, but with a _Pager_ I loose the nice colors of glow, using it like this:

     glow test.md -p | less -r

I would prefer a colored pager like **TUI** to do the same.

Issue #129, opened 2020-01-09
Missing pager functionality on Windows


muesli:
We can't use `less` on Windows for obvious reasons.

extremepayne:
...or can you?
https://github.com/bmatzelle/gow
They've ported less (among other things) to windows.
https://github.com/bmatzelle/gow/blob/master/bin/less.exe

muesli:
We could optionally support that, if we can find it in the path, but currently I'm leaning towards implementing a properly ANSI aware pager directly. `less` itself breaks in so many ways as it doesn't properly understand the sequences: try scrolling to the right and left on a small terminal, for example.

eggplants:
How about `more` or `out-host -paging`

MiMiMeowMeow:
If you install git for windows it is located in this directory:
C:\Program Files\Git\usr\bin\less.exe

However, if your set the $PAGER var to this, it chokes because of the space in "Program File"
I tried to add double qoute to the whole string but that doesn't work either.

mattcargile:
`scoop install less` works fine on windows. Colors and everything. I use the below where the important switch is `--raw-control-chars`.
```pwsh
$env:LESS = '--raw-control-chars --ignore-case --quit-if-one-screen --quiet'
```

andreynering:
After https://github.com/charmbracelet/glow/pull/679 the TUI pager will now be available with the `--tui` / `-t` flag.

PR #679, opened 2025-02-05
feat: add new `--tui` / `-t` flag

andreynering:
We had two possible paths to address this:

- Add a new `--tui` / `-t` flag
- Add a new `--mode` flag with `pager`/`tui`/`none` as possible values

Having a `--tui` flag felt a bit more reasonable to me because:

- The `-m` shortant is already taken by `--mouse` while `-t` is available. (We could of course use `-M` though).
- This avoids the need to deprecate `--pager` / `-p`.

Happy to change if you find better, though.

Let me know if you find any bugs while testing.

---

This will close the following issues:

- https://github.com/charmbracelet/glow/issues/129
- https://github.com/charmbracelet/glow/issues/489
- https://github.com/charmbracelet/glow/issues/535
- https://github.com/charmbracelet/glow/pull/548

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -144,17 +144,50 @@ func newModel(cfg Config) tea.Model {
 		cfg: cfg,
 	}
 
-	return model{
+	m := model{
 		common: &common,
 		state:  stateShowStash,
 		pager:  newPagerModel(&common),
 		stash:  newStashModel(&common),
 	}
+
+	info, err := os.Stat(cfg.Path)
+	if err != nil {
+		log.Error("unable to stat file", "file", m.common.cfg.Path, "error", err)
+		m.fatalErr = err
+		return m
+	}
+	if info.IsDir() {
+		m.state = stateShowStash
+	} else {
+		cwd, _ := os.Getwd()
+		m.state = stateShowDocument
+		m.pager.currentDocument = markdown{
+			localPath: cfg.Path,
+			Note:      stripAbsolutePath(cfg.Path, cwd),
+			Modtime:   info.ModTime(),
+		}
+	}
+
+	return m
 }
 
 func (m model) Init() tea.Cmd {
 	cmds := []tea.Cmd{m.stash.spinner.Tick}
-	cmds = append(cmds, findLocalFiles(*m.common))
+
+	switch m.state {
+	case stateShowStash:
+		cmds = append(cmds, findLocalFiles(*m.common))
+	case stateShowDocument:
+		content, err := os.ReadFile(m.common.cfg.Path)
+		if err != nil {
+			log.Error("unable to read file", "file", m.common.cfg.Path, "error", err)
+			return func() tea.Msg { return errMsg{err} }
+		}
+		body := string(utils.RemoveFrontmatter(content))
+		cmds = append(cmds, renderWithGlamour(m.pager, body))
+	}
+
 	return tea.Batch(cmds...)
 }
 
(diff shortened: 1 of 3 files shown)


ui/ui.go
```diff
@@ -144,17 +144,40 @@ func newModel(cfg Config) tea.Model {
...
 
-	return model{
+	m := model{
 		common: &common,
 		state:  stateShowStash,
 		pager:  newPagerModel(&common),
 		stash:  newStashModel(&common),
 	}
+
+	info, _ := os.Stat(cfg.Path)
```
andreynering, line 154: Let me know what I could do on an error here given this function does not accept returning `err`.

ui/ui.go
```diff
@@ -144,17 +144,40 @@ func newModel(cfg Config) tea.Model {
...
 
 func (m model) Init() tea.Cmd {
 	cmds := []tea.Cmd{m.stash.spinner.Tick}
-	cmds = append(cmds, findLocalFiles(*m.common))
+
+	switch m.state {
+	case stateShowStash:
+		cmds = append(cmds, findLocalFiles(*m.common))
+	case stateShowDocument:
+		content, _ := os.ReadFile(m.common.cfg.Path)
```
andreynering, line 176: Same here.

ui/ui.go
```diff
@@ -380,16 +403,17 @@ func waitForStatusMessageTimeout(appCtx applicationContext, t *time.Timer) tea.C
...
 		Note:      stripAbsolutePath(res.Path, cwd),
 		Modtime:   res.Info.ModTime(),
 	}
-
-	return md
 }
 
 func stripAbsolutePath(fullPath, cwd string) string {
+	if cwd == "" {
+		cwd, _ = os.Getwd()
```
raphamorim, line 415: is ok to ignore the error here right?

raphamorim (approved):
Looks fine to me, but maybe @caarlos0 @bashbunni @aymanbagabas have a better input about the error treatment

ui/ui.go
```diff
@@ -144,17 +144,40 @@ func newModel(cfg Config) tea.Model {
...
+	if info.IsDir() {
+		m.state = stateShowStash
+	} else {
+		m.state = stateShowDocument
+		m.pager.currentDocument = markdown{
+			localPath: cfg.Path,
+			Note:      stripAbsolutePath(cfg.Path, ""),
+			Modtime:   info.ModTime(),
+		}
+	}
```
aymanbagabas, line 164: Here `info` can be nil on error.
```suggestion
	info, err := os.Stat(cfg.Path)
	if err == nil {
	      if info.IsDir() {
		      m.state = stateShowStash
	      } else {
		      m.state = stateShowDocument
		      m.pager.currentDocument = markdown{
			      localPath: cfg.Path,
			      Note:      stripAbsolutePath(cfg.Path, ""),
			      Modtime:   info.ModTime(),
		      }
	      }
	}
```

ui/ui.go
aymanbagabas: I'd guard the content and return an error command

ui/ui.go
aymanbagabas: I guess it's okay to ignore it here, but I perhaps `os.Getwd` shouldn't be called here since this is a helper function

raphamorim: yea i agree, it goes out of the strip expected logic

meowgorithm:
@andreynering could you post a screenshot of the `--help` in this case? It's helpful to see what the CLI looks like with these new flags.

caarlos0 (approved):

ui/ui.go
andreynering: I moved `os.Getwd()` to elsewhere.

I also noticed we could just use the stdlib for that helper so I just changed that.

meowgorithm (approved):
LGTM. The `-t`/`--tui` flag makes sense and the way you implemented it makes sense. Nice work!

Issue #670, opened 2025-01-07
Adding a File Watcher / Live Preview to Glow


lulzshadowwalker:
First of all I would like to thank you for all of the work and efforts you put in to provide such useful, creative, and inspiring tools to the community!

## Feature Description
I believe it would be useful and a nice utility to have the preview re-rendered whenever the selected markdown file changes, especially for people who use the terminal for most of their development workflow.

## Solution Suggestions
I suggest implementing a file watcher that watches a given file for changes and efficiently re-renders the preview.

## Alternatives I Have Considered
- I tried using something like `watch glow file.md` which was *functional* but styles were messed up and lacking compared to running it normally.
- I searched through the issues on Github and noticed #135 and #560 offering a similar suggestions but which are both still open. 

## Implementation and Guidance
I am willing to implement this feature myself if you'd allow me to do so, and I would appreciate it if you could give me an insight as to what I should consider or what you have in mind for such a feature e.g. any constraints or guidelines would be appreciated. 

Thanks a lot
Will be waiting to hear from you

PR #608, opened 2024-04-16
Add reload mode

inifares23lab:
Fixes #560
Passing the -r --reload flag enables the reload mode.
This is available only when using a pager (with flag -p) and when opening one file only. 
Named it reload mode instead of watch mode because the -w flag was already taken.

Issue #560, opened 2024-02-02
Add watch mode for glow


Dangerise:
**Is your feature request related to a problem? Please describe.**
I'm using helix editor on my terminal and I need to preview my markdown file 

**Describe the solution you'd like**
Add a watch mode for glow, when the glow is watching a file, glow reload the file when the file is edited, and jump to the position which is edited.

Issue #135, opened 2020-01-15
Rerender-Mode


georgkrause:
To make it usable as preview option for markdown written in eg vim, it would be nice to have a special mode where the markdown file gets reloaded on every change. I tried running glow in watch, which didn't worked. Maybe there is already another option? If not, thats a feature request, this would be pretty neat in i3 :)

muesli:
Just a heads up: I'm currently looking into why `watch` is behaving so weirdly with glow.

krypty:
Using `entr`, it renders correctly but, like watch command, if the document is longer than your terminal height you will only see the bottom part.

`ls /path/to/your/markdown/file.md | entr -sr 'glow '`

ellisonleao:
I've started a neovim plugin using glow as a markdown preview. Unfortunately there is no hot reload feature yet, but you can already check your md file in a floating window after saving it.

You can check it out here: https://github.com/npxbr/glow.nvim

[![glow.nvim](https://i.postimg.cc/y6GX34Mq/glow.gif)](https://postimg.cc/4nzhjM4w)

pratclot:
There is also a way to do it in terminal without `watch`:
```
while true; do clear; glow README.md; sleep 2; done
```

alexzanderr:
> There is also a way to do it in terminal without `watch`:
> 
> ```
> while true; do clear; glow README.md; sleep 2; done
> ```

wow. i never thought to make my own glow live server. thanks man.

but there is a problem with `clear`, the screen is flashing very bad. how do i get rid of the flashing?

pratclot:
Hey @alexzanderr . I made this `glower.sh` script:
```
#!/bin/sh

script -q -c "glow README.md" /dev/null
```
Note that `script` has different syntax on MacOS. Then I launched this:
```
while true; do echo 12333 >> README.md ; sleep 1; done
```
And finally in the new terminal did this:
```
while true; do output=$(clear; ./glower.sh); echo "$output" | head -n $(tput lines); sleep 1; done
```
It requires me to press Enter twice before it starts working, and it outputs some garbage before and after the file content, but it does the job.

There is a very inspiring [SO question](https://unix.stackexchange.com/questions/81167/prevent-text-screen-blinking-when-doing-clear) with a similar thing, and there are two solutions. One is about `tput` tricks, another utilizes the so-called "double buffering". I could not get much with the former, but the latter worked.

The color is preserved by using `script` to call `glow`.

[...]

PR #680, opened 2025-02-06
feat: watch for file changes and reload automatically

andreynering:
We're using [fsnotify](https://github.com/fsnotify/fsnotify) for this.

This required quite some testing to get right, but I believe it's working reliably!

Let me know if you have any suggestions to improve the implementation, or if you find any bugs. I'll make a couple of comments on the code below.

---

This will close these issues:

- https://github.com/charmbracelet/glow/issues/670
- https://github.com/charmbracelet/glow/issues/135
- https://github.com/charmbracelet/glow/issues/560
- https://github.com/charmbracelet/glow/pull/608
- https://github.com/charmbracelet/glow/pull/427

### DIFF
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -14,6 +15,7 @@ import (
 	"github.com/charmbracelet/glow/v2/utils"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/charmbracelet/log"
+	"github.com/fsnotify/fsnotify"
 	runewidth "github.com/mattn/go-runewidth"
 	"github.com/muesli/reflow/ansi"
 	"github.com/muesli/reflow/truncate"
@@ -78,6 +80,7 @@ var (
 
 type (
 	contentRenderedMsg string
+	reloadMsg          struct{}
 )
 
 type pagerState int
@@ -99,6 +102,8 @@ type pagerModel struct {
 	// Current document being rendered, sans-glamour rendering. We cache
 	// it here so we can re-render it on resize.
 	currentDocument markdown
+
+	watcher *fsnotify.Watcher
 }
 
 func newPagerModel(common *commonModel) pagerModel {
@@ -107,11 +112,13 @@ func newPagerModel(common *commonModel) pagerModel {
 	vp.YPosition = 0
 	vp.HighPerformanceRendering = config.HighPerformancePager
 
-	return pagerModel{
+	m := pagerModel{
 		common:   common,
 		state:    pagerStateBrowse,
 		viewport: vp,
 	}
+	m.initWatcher()
+	return m
 }
 
 func (m *pagerModel) setSize(w, h int) {
@@ -159,6 +166,7 @@ func (m *pagerModel) showStatusMessage(msg pagerStatusMessage) tea.Cmd {
 }
 
 func (m *pagerModel) unload() {
+	log.Debug("unload")
 	if m.showHelp {
 		m.toggleHelp()
 	}
@@ -168,6 +176,7 @@ func (m *pagerModel) unload() {
 	m.state = pagerStateBrowse
 	m.viewport.SetContent("")
 	m.viewport.YOffset = 0
+	m.unwatchFile()
 }
 
 func (m pagerModel) update(msg tea.Msg) (pagerModel, tea.Cmd) {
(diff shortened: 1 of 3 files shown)


ui/ui.go
```diff
@@ -151,9 +151,13 @@ func newModel(cfg Config) tea.Model {
 		stash:  newStashModel(&common),
 	}
 
-	info, err := os.Stat(cfg.Path)
+	path := cfg.Path
+	if path == "" {
+		path = "."
+	}
+	info, err := os.Stat(path)
```
andreynering, line 158: This is an unrelated bug fix, that I believe was introduced by me on my previous PR. When no argument is given, we should default to the current directory.

ui/pager.go
```diff
@@ -451,3 +464,54 @@ func glamourRender(m pagerModel, markdown string) (string, error) {
...
+			}
+			log.Debug("fsnotify event", "file", path, "event", event.Op)
+			return reloadKeyMsg
+		case err, ok := <-m.watcher.Errors:
+			if !ok {
+				continue
+			}
+			log.Debug("fsnotify error", "file", path, "error", err)
+		}
+	}
```
andreynering, line 500: The usual way to watch for these channels would be inside a goroutine. It wasn't clear for me how to address that though, as I don't know if there is any way to send a message from a remote context (another goroutine), specially considering this channel could receive multiple messages.

Given Bubbletea _appears_ to require an inline `return` here to force an update, I had to make this loop inline inside a `cmd`. Let me know if we have a better pattern for this, or if you have a better idea on how to implement it.

andreynering:
Questions:

- Do we want a flag for this?
- If so, should it be `--watch` or `--disable-watch`?
- Do we want to watch enabled by default?

ui/pager.go
aymanbagabas: I _think_ this is fine, Bubble Tea commands run in goroutines

meowgorithm:
It looks like we're also losing the top few lines of some files in `tmux`. I haven't looked that deeply yet but it _may_ be the `log.Debug` calls in the source here.

andreynering:
@meowgorithm That's very curious. I just tried inside `tmux` but I simply couldn't reproduce it in my machine. Have you tried in the `master` branch to see if it also happens there?

andreynering:
@meowgorithm Also, what terminal app are you using?

meowgorithm:
Rio and Kitty over here. I looks like bumping Bubble Tea fixes rendering for me, but we're still missing top few lines, which is curious. I don't see that issue on `master`. I also haven't poked around the code yet, fwiw, which I will do!

ui/pager.go
```diff
@@ -451,3 +464,54 @@ func glamourRender(m pagerModel, markdown string) (string, error) {
...
+	log.Info("fsnotify watching file", "file", path)
+
+	for {
+		select {
+		case event, ok := <-m.watcher.Events:
+			if !ok || !event.Has(fsnotify.Write) {
+				continue
+			}
+			log.Debug("fsnotify event", "file", path, "event", event.Op)
+			return reloadKeyMsg
```
caarlos0, line 493: i would probably make a `type reloadMsg struct{}` instead of triggering a key press event

caarlos0 (approved):
one small comment, lgtm otherwise

meowgorithm:
> I have another unrelated request if you were ok to fix/improve: seems glow default using true color which is like '^[]11;rgb:2828/2c2c/3434^G^[[4;1R' this maybe a problem for some terminal, was it ok just using ansi color or auto-fit it on term?

@Shane-XB-Qian colors are automatically downsampled if the

@caarlos0 do you know what this semgrep check is?

meowgorithm:
> i have another unrelated request if you were ok to fix/improve:
seems glow default using true color which is like '^[]11;rgb:2828/2c2c/3434^G^[[4;1R'
this maybe a problem for some terminal, was it ok just using ansi color or auto-fit it on term?

@Shane-XB-Qian the color will automatically downsample if the terminal can't display those colors.

@caarlos0 do we know what this `semgrep` tests is all about?

meowgorithm:
> Questions:
> 
> * Do we want a flag for this?
> * If so, should it be `--watch` or `--disable-watch`?
> * Do we want to have watch enabled by default?

@andreynering For now, let's enable it by default, with no options. We can add a disable flag as the need arises.

ui/pager.go
andreynering: Good suggestion. Just pushed.

raphamorim (approved):

Shane-XB-Qian:
> > I have another unrelated request if you were ok to fix/improve: seems glow default using true color which is like '^[]11;rgb:2828/2c2c/3434^G^[[4;1R' this maybe a problem for some terminal, was it ok just using ansi color or auto-fit it on term?
> 
> @Shane-XB-Qian colors are automatically downsampled if the

not sure what you mean, but issue gone, not sure what happened, maybe behind pkg was updated.

Shane-XB-Qian:
> @andreynering commented on this pull request. > - info, err := os.Stat(cfg.Path) + path := cfg.Path + if path == "" { + path = "." + } + info, err := os.Stat(path) This is an unrelated bug fix, that I believe was introduced by me on my previous PR. When no argument is given, we should default to the current directory.
> FYI: you also made pipeline fail e.g 'cat foo.md | glow', another unrelated bug but from your last PR too.

and the stripPath function you modified at last PR, looks not well too, i raised https://github.com/charmbracelet/glow/pull/683 wish ok.

andreynering:
Thanks @Shane-XB-Qian, I'll review your PR ASAP.

Shane-XB-Qian:
pipeline for tui e.g 'cat foo.md | glow -t' failed (or didnot impl it actually), i raised #684 for it
// but not familiar with 'tea' codebase, wish no issues.

PR #683, opened 2025-02-11
fix: correct path strip

Shane-XB-Qian:
* last PR changed 'strings.ReplaceAll' to 'filepath.Rel'
* but still did not handle (or think about) symlink well


andreynering (approved):
Thanks you!

PR #684, opened 2025-02-12
fix: make pipeline for tui work

Shane-XB-Qian:
* 'cat foo.md | glow -t'


Shane-XB-Qian:
corner case when/if pipelined content was empty is going to be fixed at #686

PR #687, opened 2025-02-13
fix: disallow 'h' key when was pipeline

Shane-XB-Qian:
* 1. if pager was 'less' then 'h' was its native 'help'
* 2. if pager was 'tui' then seems no sense 'h' to back

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,4 +22,4 @@ jobs:
   ruleguard:
     uses: charmbracelet/meta/.github/workflows/ruleguard.yml@main
     with:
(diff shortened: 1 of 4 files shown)


Shane-XB-Qian:
1. test err of ruleguard, the fix had committed at #686
2. test err of coverage, seems a ci network issue, not relevant.

Shane-XB-Qian:
maybe should also disable 'h' when 'glow -t foo.md' which seems mostly for md watch,
for now in such case even emit 'h' it would not be nice back to md list (from md doc), a bit abnormal.

PR #734, opened 2025-04-02
fix: watch for write+create changes on file parent

### LINKED ISSUE #733: Live Reload Not Working With Vim/Neovim
**Describe the bug**
The live reload feature introduced in #680 only watches for writes.

Editors like Vim will create a buffer and then when the buffer is saved, replace the current version of the file with the version that was in the buffer. This gets sent as a `CREATE` and not a `WRITE` and the file watcher doesn't refresh the file.

**Setup**
 - OS [EndeavourOS 6.13.8-arch1-1/macOS]
 - Shell [fish]
 - Terminal Emulator [kitty]
 - Terminal Multiplexer [Zellij]
 - Locale [en_US.UTF-8]

**To Reproduce**
1. Open file with glow in a pager or the TUI
2. Edit file with Vim/Neovim and save
3. Markdown preview doesn't reload

**Source Code**
Please include source code if needed to reproduce the behavior. 

**Expected behavior**

Markdown preview reloads when it's written to

Ktrod:
Closes #733 

Follows the guidelines in https://github.com/fsnotify/fsnotify?tab=readme-ov-file#watching-a-file-doesnt-work-well to watch the files parent directory instead of the file itself.

Tested on both mac and Linux (Arch). On mac it seems like only the CREATE event will come through followed by a few others. On Linux it will be a CREATE followed by a WRITE. In both cases watching the parent directory makes sure any CREATE events don't have an impact on the watched file.

Might end up with more frequent reloads but I haven't dug into it that much.

### DIFF
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -476,14 +476,14 @@ func (m *pagerModel) initWatcher() {
 }
 
 func (m *pagerModel) watchFile() tea.Msg {
-	pathDir := filepath.Dir(m.currentDocument.localPath)
+	dir := m.localDir()
 
-	if err := m.watcher.Add(pathDir); err != nil {
-		log.Error("error adding file to fsnotify watcher", "error", err)
+	if err := m.watcher.Add(dir); err != nil {
+		log.Error("error adding dir to fsnotify watcher", "error", err)
 		return nil
 	}
 
-	log.Info("fsnotify watching file", "file", pathDir)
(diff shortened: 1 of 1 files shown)


andreynering:
Thank you very much @Ktrod!

I was able to reproduce the issue with [Helix](https://helix-editor.com/), and your fix work correctly.

I pushed a commit with small code changes before merging.

PR #735, opened 2025-04-02
chore: do some basic repo maintenance (editorconfig, linting, etc)

andreynering:


.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+[*]
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 2
+
+[*.go]
+indent_style = tab
+indent_size = 8
```
raphamorim, line 14: 8 tabs?

.github/workflows/lint-sync.yml
```diff
@@ -0,0 +1,13 @@
+name: lint-sync
+on:
+  schedule:
+    - cron: "0 0 * * 0"
```
raphamorim, line 4: could you add a comment marking what time ref is that?

.golangci.yml
```diff
@@ -1,33 +1,49 @@
...
     - whitespace
+    - wrapcheck
+  exclusions:
+    generated: lax
+    presets:
+      - common-false-positives
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
ccoVeille, line 36: These ones are added blindly by `golangci-lint migrate`

Abd it's normal as they were previous excluded by the v1 via default configuration

You don't have such folder, so you should remove them


```suggestion
```

.golangci.yml
```diff
@@ -1,33 +1,49 @@
...
+formatters:
+  enable:
+    - gofumpt
+    - goimports
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
ccoVeille, line 49: Same as https://github.com/charmbracelet/glow/pull/735#discussion_r2027136655

```suggestion
```

config_cmd.go
```diff
@@ -39,13 +39,13 @@ var configCmd = &cobra.Command{
 
 		c, err := editor.Cmd("Glow", configFile)
 		if err != nil {
-			return err
+			return fmt.Errorf("glow: unable to set config file: %w", err)
```
ccoVeille, line 42: You are wrapping all error, it's great!

But you are adding the "glow: " prefix everywhere.

It might lead to error being reported like this

"glow: foo bar: glow quz ...: glow: ..."

I think you should consider removing all the "glow: " prefix, and adding it only in a few place in main.go

Like when calling Execute in RunE, if you have an error, you wrap it to include "glow: "

You might have to do it also when validating the config in PersitantPreRunE

What do you think about it?

utils/utils.go
```diff
@@ -30,7 +31,7 @@ func detectFrontmatter(c []byte) []int {
 	return []int{-1, -1}
 }
 
-// Expands tilde and all environment variables from the given path.
+// ExpandPath tilde and all environment variables from the given path.
```
bashbunni, line 34:
```suggestion
// ExpandPath expands tilde and all environment variables from the given path.
```

bashbunni (approved):
looks good to me

config_cmd.go
andreynering: That's a good point, and one of the reasons I think enforcing wraps in 100% os cases does not necessarily makes sense.

I removed the `glow:` prefix for now.

viZi67 (approved):
H

PR #765, opened 2025-05-13
Add the ability to sort by date or title

### LINKED ISSUE #228: Different sorting options
It would be helpful to be able to organize documents in several different ways. I would have command line options and config options to allow:

* sorting by creation date (current method)
* sorting by title

I can't think of any others at the moment, but I'm sure other people might have sorting methods they'd like

cameron-michie:
* Closes #228

### Describe your changes
Adds the ability to press 'y' to toggle the sort by date. Also adds the ability to press 't' to switch back to sorting by title, and can sort reverse alphabetically by pressing repeatedly. These options now appear in the fullView after pressing '?'.

### Related issue/discussion: https://github.com/charmbracelet/glow/issues/228

### DIFF
--- a/ui/sort.go
+++ b/ui/sort.go
@@ -6,59 +6,35 @@ import (
 	"time"
 )
 
-type SortType int
-type SortOrder int
+type sortOrder bool
 
 const (
-	SortByNote SortType = iota
-	SortByDate
-	SortByTitle
+	ascending  sortOrder = true
+	descending sortOrder = false
 )
 
-const (
-	SortAscending SortOrder = iota
-	SortDescending
-)
-
-type SortState struct {
-	Type  SortType
-	Order SortOrder
-}
-
-func (s *SortState) Toggle(newType SortType) {
-	if s.Type == newType {
-		// Toggle order if same type
-		if s.Order == SortAscending {
-			s.Order = SortDescending
+func sortMarkdowns(mds []*markdown, byDate bool, ascending sortOrder) {
+	if byDate {
+		if ascending {
+			slices.SortStableFunc(mds, func(a, b *markdown) int {
+				return compareTime(a.Modtime, b.Modtime)
+			})
 		} else {
-			s.Order = SortAscending
+			slices.SortStableFunc(mds, func(a, b *markdown) int {
+				return -compareTime(a.Modtime, b.Modtime)
+			})
 		}
 	} else {
-		// Set new type with default ascending order
-		s.Type = newType
-		s.Order = SortAscending
+		if ascending {
+			slices.SortStableFunc(mds, func(a, b *markdown) int {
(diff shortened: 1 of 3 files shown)


ui/sort.go
```diff
@@ -3,10 +3,47 @@ package ui
...
+			slices.SortStableFunc(mds, func(a, b *markdown) int {
+				return cmp.Compare(a.Note, b.Note)
+			})
+		} else {
+			slices.SortStableFunc(mds, func(a, b *markdown) int {
+				return -cmp.Compare(a.Note, b.Note)
+			})
+		}
+	}
+}
```
andreynering, line 38: Some suggestions:

1. Instead of repeating `slices.SortStableFunc` 4 times, assigning functions and calling  in the end (make it a one-liner).
2. Change `if/else` into a `switch` case to reduce block inside block.

```suggestion
func sortMarkdowns(mds []*markdown, sortByDate bool, order sortOrder) {
	var fn func(a, b *markdown) int

	switch {
	case sortByDate && order == ascending:
		fn = func(a, b *markdown) int { return compareTime(a.Modtime, b.Modtime) }
	case sortByDate && order == descending:
		fn = func(a, b *markdown) int { return -compareTime(a.Modtime, b.Modtime) }
	case !sortByDate && order == ascending:
		fn = func(a, b *markdown) int { return cmp.Compare(a.Note, b.Note) }
	case !sortByDate && order == descending:
		fn = func(a, b *markdown) int { return -cmp.Compare(a.Note, b.Note) }
	}

	slices.SortStableFunc(mds, fn)
}
```

andreynering (commented):
Hi @cameron-michie,

I made a small suggestion.

ui/sort.go
mabhub: _m2¢_:


```go
func sortMarkdowns(mds []*markdown, sortByDate bool, order sortOrder) {
    cmpFn := func(a, b *markdown) int { return cmp.Compare(a.Note, b.Note) }
    if sortByDate {
        cmpFn = func(a, b *markdown) int { return compareTime(a.Modtime, b.Modtime) }
    }
    if order == descending {
        orig := cmpFn
        cmpFn = func(a, b *markdown) int { return -orig(a, b) }
    }
    slices.SortStableFunc(mds, cmpFn)
}
```

mabhub:
Thanks for working on this!

One concern: the change in `addMarkdowns` removes the `len(mds) == 0` guard and unconditionally overwrites `filteredMarkdowns` with `markdowns`. This means that if a user has an active filter and new local files come in, the filtered view gets silently reset. I think the original guards were there for a reason.

liamasis:
I am also interested in this, I use this at work and have implemented my own at home, but would need a stable release to download this at work :)

Issue #766, opened 2025-05-14
Bug: Trying to edit with no files present will crash glow


MarkusBillharz:
**Describe the bug**
When no files are present, pressing e will crash glow.

**To Reproduce**
1. Go to an empty directory
2. Open glow
3. Press e
4. See crash

**Additional context**
```
Caught panic:

runtime error: invalid memory address or nil pointer dereference

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.24.0/x64/src/runtime/debug/stack.go:26 +0x5e
runtime/debug.PrintStack()
        /opt/hostedtoolcache/go/1.24.0/x64/src/runtime/debug/stack.go:18 +0x13
github.com/charmbracelet/bubbletea.(*Program).recoverFromPanic(0xc0004ecdc0)
        /home/runner/go/pkg/mod/github.com/charmbracelet/bubbletea@v1.3.3/tea.go:733 +0x8b
panic({0x1374be0?, 0x1c524f0?})
        /opt/hostedtoolcache/go/1.24.0/x64/src/runtime/panic.go:787 +0x132
github.com/charmbracelet/glow/v2/ui.(*stashModel).handleDocumentBrowsing(0xc000234008, {0x13db4c0, 0xc0004c7200?})
        /home/runner/go/pkg/mod/github.com/charmbracelet/glow/v2@v2.1.0/ui/stash.go:521 +0x37f
github.com/charmbracelet/glow/v2/ui.stashModel.update({0xc00036ec80, {0x0, 0x0}, {{{0x1c6a700, 0x4, 0x4}, 0x5f5e100}, {0x1c8c640, 0x400, {0x0, ...}, ...}, ...}, ...}, ...)
        /home/runner/go/pkg/mod/github.com/charmbracelet/glow/v2@v2.1.0/ui/stash.go:449 +0x53b
github.com/charmbracelet/glow/v2/ui.model.Update({0xc00036ec80, 0x0, {0x0, 0x0}, {0xc00036ec80, {0x0, 0x0}, {{{...}, 0x5f5e100}, {0x1c8c640, ...}, ...}, ...}, ...}, ...)
        /home/runner/go/pkg/mod/github.com/charmbracelet/glow/v2@v2.1.0/ui/ui.go:313 +0x1035
[...]

PR #767, opened 2025-05-14
Fix: prevent crash on edit attempt with no files present

MarkusBillharz:
Hello,
currently, attempting to edit with no files present causes the app to crash because m.selectedMarkdown() returns nil, which leads to a runtime error when it is passed to openEditor().

#### Changes:
- [handle case when selected md is nil, which led to the crash](https://github.com/charmbracelet/glow/commit/b49e1931aaf643ca0aba76564e7c8dc33d721178) 
(tea.Batch seems to handle it gracefully now.)
- [only show edit help if documents are available](https://github.com/charmbracelet/glow/commit/2a4725cade2d9649f932b6e199b458551b78fcb7)

New empty state:
empty
Documents present:
filled

#### This will close https://github.com/charmbracelet/glow/issues/766

I also created a [version](https://github.com/MarkusBillharz/glow/compare/task/fix_edit_crash_on_empty_stash...MarkusBillharz:glow:task/create_file_if_stash_is_empty) on top of this fix that points to a default file instead, but wasn’t sure if that aligns with your project’s goals.

Thanks!~

### DIFF
--- a/ui/stash.go
+++ b/ui/stash.go
@@ -518,6 +518,12 @@ func (m *stashModel) handleDocumentBrowsing(msg tea.Msg) tea.Cmd {
 		// Edit document in EDITOR
 		case "e":
 			md := m.selectedMarkdown()
+
+			// In case no file is available
+			if md == nil {
+				return nil
+			}
+
 			return openEditor(md.localPath, 0)
 
 		// Open document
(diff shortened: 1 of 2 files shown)


meowgorithm (approved):
What a fantastic PR. Thanks for the excellent fix and thoughtful, thorough commentary.

meowgorithm:
The other version where it creates a file if no files are found is very smart, but I say let's hold off for now as I think it will open up into bigger decisions such as whether one should be able to create new files in general, even if a README is present.

PR #786, opened 2025-06-23
feat: use fang

caarlos0:
using [fang](https://github.com/charmbracelet/fang):

CleanShot 2025-06-23 at 15 48 39@2x


main.go
```diff
@@ -374,7 +374,7 @@ func main() {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	if err := rootCmd.Execute(); err != nil {
+	if err := fang.Execute(context.Background(), rootCmd); err != nil {
```
ccoVeille, line 377: Did you consider using this?

https://pkg.go.dev/os/signal#NotifyContext

aymanbagabas (approved):

main.go
```diff
@@ -374,7 +374,7 @@ func main() {
...
 		_ = closer()
```
ccoVeille, line 378: Why not using a defer after setting up the logger?

main.go
```diff
@@ -44,11 +46,9 @@ var (
...
-		Use:   "glow [SOURCE|DIR]",
-		Short: "Render markdown on the CLI, with pizzazz!",
-		Long: paragraph(
-			fmt.Sprintf("\nRender markdown on the CLI, %s!", keyword("with pizzazz")),
-		),
+		Use:              "glow [SOURCE|DIR]",
+		Short:            "Render markdown on the CLI, with pizzazz!",
+		Long:             fmt.Sprintf("Render markdown on the CLI, %s!", keyword("with pizzazz")),
 		SilenceErrors:    false,
 		SilenceUsage:     true,
```
ccoVeille, line 53: Couldn't fang.Execute sets this two?

BTW, do you really want SilentErrors to be false?

ccoVeille (commented):
Some  questions

andreynering (approved):

ccoVeille:
@caarlos0 could you take a look at my comments?

main.go
caarlos0: we could probably do that in fang I think 🤔 

something like `fang.WithNotifyContext(SIGINT, SIGTERM, etc...)` 🤔

caarlos0: https://github.com/charmbracelet/fang/pull/48

main.go
caarlos0: it does, should remove it

main.go
caarlos0: if my memory serves me, `os.Exit(1)` skips the `defer` calls 🤔

main.go
andreynering: Is `os/signal.Notify` really useful here if we're not doing anything special?

The default Go behavior is to finish the app, and that's basically what we want here.

Added some comments on https://github.com/charmbracelet/fang/pull/48/files#r2172861345. Let me know if I'm missing anything.

caarlos0: > The default Go behavior is to finish the app, and that's basically what we want here.

that's right, we should be using the context on the tea app as well - just pushed a fix for that 

ideally then, the app should gracefully quit as the context is canceled, and any other operations can be canceled as well (e.g. exec.Commands, http requests, etc)

ccoVeille (approved):
I love seeing context everywhere 🤩

meowgorithm:
@caarlos0 are we good with this one? Is there anything else from @ccoVeille which needs to be addressed?

main.go
ccoVeille: Do you think there is a need to refactor then?

The question is for @caarlos0 @meowgorithm

url_test.go
```diff
@@ -15,7 +18,7 @@ func TestURLParser(t *testing.T) {
 	} {
 		t.Run(path, func(t *testing.T) {
 			t.Skip("test uses network, sometimes fails for no reason")
-			got, err := readmeURL(path)
+			got, err := readmeURL(context.Background(), path)
```
ccoVeille, line 21: This maybe as you are on Go 1.24

```suggestion
			got, err := readmeURL(t.Context(), path)
```

main.go
ccoVeille: I think we are good here

kujtimiihoxha:
@caarlos0 needs a rebase/merge

PR #800, opened 2025-07-24
Adding badge for https://repomapr.com/charmbracelet/glow

JetsettingJames:
See live system diagram + understand how each node works
Query chat to better understand repo
Repo owners get access to query analytics


caarlos0:
Hey!

That's nice, but in the case of glow, I don't see how it can be useful.

Thanks for the PR, though!
Cheers.

PR #801, opened 2025-07-24
Add support for stream input

davit312:
### Add support for stream

Currently glow waits input stream be closed and after that renders full content. With this feature input could be rendered to markdown in real time, added `--flow` flag to enable this mode.

This feature could be used with LLM's output
ex. `ollama run gemma3:4b "Create list of 5 fruits" | glow --flow`

**Important** If same text created multiple lines, then use `-w` flag to specify terminal width, glow  couldn't automatically detect terminal size and set it to default value 80 or 120 which could be bigger than actual terminal size. 

### Limitations in this mode
* no code highlighting
* no styles 
* not supports tui and pager

### Checklist before requesting a review

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my code

### If this is a feature

- [ ] I have created a discussion
- [ ] A project maintainer has approved this feature request. Link to comment:

### DIFF
--- a/flow.go
+++ b/flow.go
@@ -1,0 +1,142 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/charmbracelet/glamour"
+	"github.com/charmbracelet/lipgloss"
+	"github.com/spf13/cobra"
+)
+
+func executeWithFlow(cmd *cobra.Command, src *source, w io.Writer) error {
+	var line string
+	var nextLine string
+	for {
+		b := make([]byte, 1024)
+		n, err := src.reader.Read(b)
+		if err != nil {
+			if err == io.EOF {
+				if n == 0 && nextLine == "" {
+					return nil
+				}
+			} else {
+				return fmt.Errorf("unable to read from reader: %w", err)
+			}
+		}
+
+		if n == 0 && nextLine == "" {
+			time.Sleep(100 * time.Millisecond)
+			continue
+		}
+
+		// if buffered nextlin exists, update current line
+		if nextLine != "" {
+			line = strings.TrimLeft(nextLine, "\n")
+			if strings.HasPrefix(nextLine, "\n") {
+				fmt.Fprint(w, "\n")
+			}
+			nextLine = ""
+		}
+
+		chunk := strings.ReplaceAll(string(b[:n]), "\r", "")
+		fullLine := line + chunk
+
+		if i := strings.IndexAny(fullLine, "\n"); i != -1 {
+			line = fullLine[:i+1]
+			if len(fullLine) > i+1 {
+				nextLine = fullLine[i+1:]
+			}
+		} else {
+			line = fullLine
+		}
+
+		if strings.HasSuffix(line, "\n") {
+			fmt.Fprint(w, "\r"+line)
+			line = ""
+			continue
+		}
+		if len(line) > int(width) {
+			continue
+		}
+		fmt.Fprint(w, "\r"+line)
+	}
+}
+
+func renderString(in string) string {
+	// TODO: Implement flow rendering logic
+
+	r, err := glamour.NewTermRenderer(
+		glamour.WithColorProfile(lipgloss.ColorProfile()),
+		glamour.WithAutoStyle(),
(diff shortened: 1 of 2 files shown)


bigbitbox:
will this be merge? I am expecting this feature when I use `aichat`

munim:
Any update on this PR?  Would be really nice to have streaming functionality

anthonyrisinger:
I have a working PR I was preparing for final submission, and then it occurred to me that I should probably check if anyone else has already submitted one; the version I have drives glamour in such a way that it produces byte-identical output in all cases—providing you add a large enough `--buffer` to handle the resolution of link refs **you haven't seen yet**—and is used in all cases except TUI (which it probably could be, but I skipped for safety). It accumulates incoming data asynchronously and then finds the largest safe offset to pass to glamour (maximizing the likelihood that links get resolved). This mechanism defaults to `--buffer=-1` for unlimited, and this results in the current behavior today by accumulating the entire input before passing to glamour. `--buffer=0` is as close to unbuffered as possible, finalizing blocks the moment it's safe to do so, which in my testing is virtually all cases except unresolved links (which are almost certainly going to be left unresolved in the unbuffered case). Finally, anything `--buffer=N` create a hard boundary where hitting it means the currently accumulated block is flushed. The implementation properly supports SIGTERM and SIGINT, and is confirmed to flush properly when they are encountered. It was largely inspired by https://willmcgugan.github.io/streaming-markdown/; all this is to say I'm likely going to post it in the coming day or two unless something moves here (and I also wanted to note the edge cases and signal concerns that must be dealt with for a proper streaming implementation).

anthonyrisinger:
I don't have a preference for what actually lands, but I would like to have the features I mentioned above supported, which is proper signal handling, and the ability to tune the sliding window. In a streaming context, the output of glamour changes depending on what it's **aware** of, so the only way to do the right thing in **all** cases is to allow the user to change the size of that same window. I also like this path because it ensures the streaming implementation itself is the only path in the code in all cases; `--buffer` alone is what effectively controls whether or not there are 'markdown differences' in the final output. It grants users the convenience of streaming by default and handling unlimited size files at some point in the future—should glow change the default to e.g. `--buffer=131072` (128KiB) or something reasonable, instead of `--buffer=-1`—all with a very low chance of unresolved links, thanks to the buffering. The user experience could be further improved by adding a small timer that auto flushes the most recent block if it's connected directly to a terminal after e.g. 0.2s, instead of waiting for the complete buffer to fill; all of these behaviors match existing UNIX tools and how they deal with tty vs. pipe for the benefit of usability.

caarlos0:
sorry I haven't been able to review this yet. @anthonyrisinger your idea seems pretty good to, if you want to push it and reference it here, will make an effort to look at and test both this week 🙏🏻 

thanks and sorry for the delay ❤️

anthonyrisinger:
@caarlos0 @davit312 See https://github.com/charmbracelet/glow/pull/823 and LMK; there's a handful of outstanding issues as I ran a bit short on time, but I wanted to get it posted early too, so I left it as draft. I'm hoping we can rally the various streaming-related issues around this one; it looks like at least 2-3+ issues will close if this patch lands (possibly more because it adds some process-level quality-of-life things like signal handling). In the spirit of this PR though, I adapted mine to use `--flow`; the battery of tests *should* run against this implementation too, should you decide to pursue that path, though I'm hoping we can concentrate effort on one implementation ☺️

anthonyrisinger:
@davit312 If you're good with consolidating effort on #823 (and thus willing to close this one of your own volition) I'll begin the process of converting the most important of the 60+ shell tests into real go tests; I was going to punt on that for the PR as it's some work, but there's a ton of edge cases in there, and it'd be unfortunate to either delete or leave in shell form.

caarlos0:
After talking about this with the team, we decided can't move forward with this approach at this time. 
We have another thing in the works that would make this simpler to implement, but we don't have a timeline for it yet.
That said, thanks for bringing this up, and thanks for all the work you put into it.
Again, sorry for not reviewing this earlier, we're a rather small team and sometimes we fail to keep up with everything.

PR #823, opened 2025-09-12
feat: add high-performance streaming markdown renderer

anthonyrisinger:
> Ideas from https://willmcgugan.github.io/streaming-markdown/ inspired and influenced this work

This release introduces a new streaming architecture that processes markdown incrementally, enabling better performance and reliability for large documents and real-time scenarios.

Key improvements:

- New streaming mode with configurable flow control for different use cases
- Proper frontmatter handling - streaming documents with YAML frontmatter processes correctly without corruption or memory issues
- Enhanced nested code block support - complex documents with mixed fence levels (e.g. triple/quadruple backticks) now parse correctly
- Robust signal handling - Ctrl+C (SIGINT) and SIGTERM now terminate cleanly, preserving partial output without corruption
- Graceful pipe handling - broken pipes and downstream shutdowns handled without errors or crashes
- Immediate responsiveness for log tailing and live documentation scenarios
- Memory-efficient processing of large documents and infinite streams
- Comprehensive edge case handling including binary data, malformed UTF-8, and deeply nested markdown structures

New scenarios enabled:

- Live documentation updates and git log streaming
- Log tailing with markdown formatting
- Large single-line inputs and massive documents
- Complex nested code examples and technical documentation
- Rapid incremental updates and mixed content streams
- Robust empty or minimal input handling
- Advanced table parsing with alignment and escaped characters

All changes maintain complete backward compatibility. Existing workflows, scripts, and integrations will continue to work while also benefiting from enhanced performance and reliability. Output formatting remains identical.

Run `flow/t/test.sh` to run all the tests. There's some 60+ that pressure test from all kinds of different angles. Probably you'll want to either drop this or convert it to proper go tests later, but I got tired of repeating commands, and I wasn't quite at the point of go tests yet, so this is where I ended up 😅; use GLOW_TEST_LOG=1 if anything fails:

```bash
flow/t/test.sh
```

```markdown
🎯🎯🎯 THE 'YAAAY! GREEN SMILES' GLOW FLOW VALIDATION SUITE 🎯🎯🎯

> go test ./...

    ⏳ go test (100.00 CPU over 0m0.311s)

> flow/t/test.sh flow/t/test_arch.sh ...

    ⏳ t/test.sh flow/t/test_arch.sh (82.71 CPU over 0m5.007s)
    🎯 t/test.sh flow/t/test_arch.sh (16/16 tests passed)

> flow/t/test.sh flow/t/test_comprehensive.sh ...

    ⏳ t/test.sh flow/t/test_comprehensive.sh (13.66 CPU over 0m11.857s)
    🎯 t/test.sh flow/t/test_comprehensive.sh (25/25 tests passed)

> flow/t/test.sh flow/t/test_critical.sh ...

    ⏳ t/test.sh flow/t/test_critical.sh (100.00 CPU over 0m1.704s)
    🎯 t/test.sh flow/t/test_critical.sh (10/10 tests passed)

> flow/t/test.sh flow/t/test_edges.sh ...

    ⏳ t/test.sh flow/t/test_edges.sh (100.00 CPU over 0m0.521s)
    🎯 t/test.sh flow/t/test_edges.sh (8/8 tests passed)

> flow/t/test.sh flow/t/test_fences.sh ...

    ⏳ t/test.sh flow/t/test_fences.sh (100.00 CPU over 0m0.424s)
    🎯 t/test.sh flow/t/test_fences.sh (12/12 tests passed)

> flow/t/test.sh flow/t/test_frontmatter.sh ...

    ⏳ t/test.sh flow/t/test_frontmatter.sh (100.00 CPU over 0m0.479s)
    🎯 t/test.sh flow/t/test_frontmatter.sh (10/10 tests passed)

> flow/t/test.sh flow/t/test_hardening.sh ...

    ⏳ t/test.sh flow/t/test_hardening.sh (100.00 CPU over 0m6.685s)
    🎯 t/test.sh flow/t/test_hardening.sh (8/8 tests passed)

> flow/t/test.sh flow/t/test_pathological.sh ...

    ⏳ t/test.sh flow/t/test_pathological.sh (71.79 CPU over 0m3.291s)
    🎯 t/test.sh flow/t/test_pathological.sh (8/8 tests passed)

> flow/t/test.sh flow/t/test_progressive.sh ...

    ⏳ t/test.sh flow/t/test_progressive.sh (9.33 CPU over 0m16.363s)
    🎯 t/test.sh flow/t/test_progressive.sh (5/5 tests passed)

> flow/t/test.sh flow/t/test_stream.sh ...

    ⏳ t/test.sh flow/t/test_stream.sh (96.40 CPU over 0m5.758s)
    🎯 t/test.sh flow/t/test_stream.sh (19/19 tests passed)

> flow/t/test.sh flow/t/test_ux.sh ...

    ⏳ t/test.sh flow/t/test_ux.sh (53.86 CPU over 0m4.249s)
    🎯 t/test.sh flow/t/test_ux.sh (12/12 tests passed)

  ⏱️ t/test.sh (57.43 CPU over 0m56.371s)

🎉🎉🎉 YAAAY! ALL 133 TESTS PASSED! 🎉🎉🎉
```

There's almost certainly some duplication and maybe even some slop, but I curated it pretty hard. Let me know what to trim or change and we can do it ☺️

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -10,12 +11,15 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/caarlos0/env/v11"
 	"github.com/charmbracelet/glamour"
 	"github.com/charmbracelet/glamour/styles"
+	"github.com/charmbracelet/glow/v2/flow"
 	"github.com/charmbracelet/glow/v2/ui"
 	"github.com/charmbracelet/glow/v2/utils"
 	"github.com/charmbracelet/lipgloss"
@@ -25,6 +29,37 @@ import (
 	"github.com/spf13/viper"
 	"golang.org/x/term"
 )
+
+const (
+	// ExitCodeTimeout is the exit code for timeout (used by timeout command)
+	ExitCodeTimeout = 124
+
+	// ExitCodeSIGPIPE is the exit code for SIGPIPE (when downstream closes)
+	ExitCodeSIGPIPE = 141
+
+	// ExitCodeSIGINT is the signal offset for SIGINT (Ctrl+C)
+	ExitCodeSIGINT = 128 + 2
+
+	// ExitCodeSIGTERM is the signal offset for SIGTERM
+	ExitCodeSIGTERM = 128 + 15
+)
+
+// ExitError represents an error that should cause the program to exit with a specific code
+type ExitError struct {
+	Code int
+	Err  error
+}
+
+func (e *ExitError) Error() string {
+	if e.Err != nil {
+		return e.Err.Error()
+	}
+	return fmt.Sprintf("exit with code %d", e.Code)
+}
+
+func (e *ExitError) Unwrap() error {
+	return e.Err
+}
 
 var (
 	// Version as provided by goreleaser.
@@ -36,6 +71,7 @@ var (
 	configFile       string
 	pager            bool
 	tui              bool
+	window           int64
 	style            string
 	width            uint
 	showAllFiles     bool
@@ -168,6 +204,7 @@ func validateOptions(cmd *cobra.Command) error {
 	mouse = viper.GetBool("mouse")
 	pager = viper.GetBool("pager")
 	tui = viper.GetBool("tui")
+	window = viper.GetInt64("flow")
 	showAllFiles = viper.GetBool("all")
 	preserveNewLines = viper.GetBool("preserveNewLines")
 	showLineNumbers = viper.GetBool("showLineNumbers")
@@ -197,9 +234,6 @@ func validateOptions(cmd *cobra.Command) error {
 				width = uint(w) //nolint:gosec
 			}
 
-			if width > 120 {
-				width = 120
-			}
 		}
 		if width == 0 {
 			width = 80
@@ -226,13 +260,17 @@ func execute(cmd *cobra.Command, args []string) error {
 		return err
 	} else if yes {
 		src := &source{reader: os.Stdin}
-		defer src.reader.Close() //nolint:errcheck
+		// DO NOT close stdin - it's not ours to close and causes race conditions
 		return executeCLI(cmd, src, os.Stdout)
 	}
 
 	switch len(args) {
 	// TUI running on cwd
 	case 0:
+		if cmd.Flags().Changed("flow") {
+			// --flow flag was explicitly provided, incompatible with directories
+			return fmt.Errorf("--flow is not compatible with directory input")
+		}
 		return runTUI("", "")
 
 	// TUI with possible dir argument
@@ -241,11 +279,19 @@ func execute(cmd *cobra.Command, args []string) error {
 		// an argument to the non-TUI version of Glow (via fallthrough).
 		info, err := os.Stat(args[0])
 		if err == nil && info.IsDir() {
-			p, err := filepath.Abs(args[0])
-			if err == nil {
-				return runTUI(p, "")
+			// Directory input detected
+			if cmd.Flags().Changed("flow") {
+				// --flow flag was explicitly provided, incompatible with directories
+				return fmt.Errorf("--flow is not compatible with directory input")
 			}
+			// No --flow specified, launch TUI for directory
+			p, err := filepath.Abs(args[0])
+			if err != nil {
+				return fmt.Errorf("could not resolve directory path: %w", err)
+			}
+			return runTUI(p, "")
 		}
+		// Not a directory, treat as regular file argument
 		fallthrough
 
 	// CLI
@@ -266,17 +312,14 @@ func executeArg(cmd *cobra.Command, arg string, w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	defer src.reader.Close() //nolint:errcheck
+	// Only close if it's not stdin (stdin is not ours to close)
+	if src.reader != os.Stdin {
+		defer src.reader.Close() //nolint:errcheck
+	}
 	return executeCLI(cmd, src, w)
 }
 
 func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
-	b, err := io.ReadAll(src.reader)
-	if err != nil {
-		return fmt.Errorf("unable to read from reader: %w", err)
-	}
-
-	b = utils.RemoveFrontmatter(b)
 
 	// render
 	var baseURL string
@@ -300,17 +343,6 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
 		return fmt.Errorf("unable to create renderer: %w", err)
 	}
 
-	content := string(b)
-	ext := filepath.Ext(src.URL)
-	if isCode {
-		content = utils.WrapCodeBlock(string(b), ext)
-	}
-
-	out, err := r.Render(content)
-	if err != nil {
-		return fmt.Errorf("unable to render markdown: %w", err)
-	}
-
 	// display
 	switch {
 	case pager || cmd.Flags().Changed("pager"):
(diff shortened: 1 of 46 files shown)


main.go
```diff
@@ -300,45 +320,47 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
-	if err != nil {
-		return fmt.Errorf("unable to render markdown: %w", err)
-	}
-
-	// display
-	switch {
-	case pager || cmd.Flags().Changed("pager"):
-		pagerCmd := os.Getenv("PAGER")
-		if pagerCmd == "" {
-			pagerCmd = "less -r"
```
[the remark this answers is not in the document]
anthonyrisinger: (Should have auto resolved, but I didn't submit the review)

main.go
```diff
@@ -300,45 +320,47 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
-		c.Stdin = strings.NewReader(out)
-		c.Stdout = os.Stdout
-		if err := c.Run(); err != nil {
-			return fmt.Errorf("unable to run command: %w", err)
-		}
-		return nil
-	case tui || cmd.Flags().Changed("tui"):
-		path := ""
-		if !isURL(src.URL) {
-			path = src.URL
```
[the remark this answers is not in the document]
anthonyrisinger: (Should have auto resolved, but I didn't submit the review)

main.go
```diff
@@ -300,45 +323,99 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
+		// Send buffered output to pager
 		pagerCmd := os.Getenv("PAGER")
 		if pagerCmd == "" {
 			pagerCmd = "less -r"
 		}
-
 		pa := strings.Split(pagerCmd, " ")
 		c := exec.Command(pa[0], pa[1:]...) //nolint:gosec
-		c.Stdin = strings.NewReader(out)
+		c.Stdin = bytes.NewReader(buf.Bytes())
```
anthonyrisinger, line 366: We should be streaming to the PAGER as well.

main.go
```diff
@@ -300,45 +323,99 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
+
+		rendered, _ := r.RenderBytes(content)
+		return rendered
 	}
 
-	// display
+	// Determine output mode and handle accordingly
 	switch {
 	case pager || cmd.Flags().Changed("pager"):
+		// Buffer output for pager
```
anthonyrisinger, line 352: We should be streaming to the PAGER as well.

main.go
```diff
@@ -300,45 +323,99 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
-	}
+	// Stream with incremental rendering using context
+	renderFunc := func(content []byte) []byte {
+		// Detect truly pathological patterns that will hang glamour
+		contentLen := len(content)
 
-	out, err := r.Render(content)
-	if err != nil {
-		return fmt.Errorf("unable to render markdown: %w", err)
+		// Check for excessive unclosed brackets that cause O(n²) parsing
```
anthonyrisinger, line 331: I'm not sure if this is worth doing; I'll do some investigation here and vet this part a bit more before dropping DRAFT.

main.go
```diff
@@ -300,45 +323,99 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
+		if contentLen > 10000 {
+			// Count unclosed brackets - these cause exponential parsing time
+			openBrackets := bytes.Count(content, []byte("["))
+			closeBrackets := bytes.Count(content, []byte("]"))
+			unclosed := openBrackets - closeBrackets
+
+			// If more than 10% unclosed brackets in large content, skip rendering
+			if unclosed > contentLen/10 {
```
anthonyrisinger, line 339: This is arbitrary; if reasonable, include reasoning.

flow.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+}
+
+// shouldFlush determines if buffer should be flushed based on flow config
+func (fb *FlowBuffer) shouldFlush() bool {
+	if fb.config.Size == 0 {
+		// No flow: never flush until EOF
+		return false
+	}
+	if fb.config.Size < 0 {
```
anthonyrisinger, line 215: I think we could have our cake and eat it too if we interpreted negative sizes as 'aggressive flush first half of double max buffer', i.e. similar to incremental `--flow=N` buffering, but allowing some early opportunistic flushing, maybe we look for two safe boundaries instead of just one or whatever. Ultimately, I think I'm trying to ensure a good user experience for things like long code blocks, but we might need better glamour support or more state tracking so we can indent when the buffering is too small to see the entire code block. This might also enable incremental rendering of things _within_ the code, which is a good thing, but having better support for incremental **anything** in glamour proper would be much nicer, e.g. maybe a few new `Options()` to help manage newlines, large code blocks, large tables, and other large/nested structures we'd like to incrementally render as they come in knowing full well, they are nested within something else. Glamour might return some sort of object that identifies which structure were currently nested in, and then we just continue currying whatever object Glamour returns (it could be a totally black box thing from the consuming API point of view, akin to a cursor).

anthonyrisinger:
There's [always 🫠] more cleanup that can be done, but it should be pretty solid now. I added additional tests to ensure progressive rendering and flushing after finding a few memory-leak type edge cases. I'm going to continue hardening it for another day or so; it'll be completely ready before next week starts anew. If I can get the nod that we'll consolidate efforts here, I'll translate the exhaustive and high-value shell tests to proper go tests and weed out any duplicates (I'll do whatever's preferred here, even more so once I know it will land if I do).

main.go
```diff
@@ -271,12 +278,28 @@ func executeArg(cmd *cobra.Command, arg string, w io.Writer) error {
 }
 
 func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
-	b, err := io.ReadAll(src.reader)
-	if err != nil {
-		return fmt.Errorf("unable to read from reader: %w", err)
-	}
-
-	b = utils.RemoveFrontmatter(b)
+	// Create context with signal handling for graceful shutdown
```
anthonyrisinger, line 281: Most/all signal handling dance needs to be moved to the very top `main` function.

main.go
```diff
@@ -300,45 +323,99 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
-		if _, err = fmt.Fprint(w, out); err != nil {
-			return fmt.Errorf("unable to write to writer: %w", err)
+		// Stream directly to output (preserves streaming performance)
+		// Wrap stdout in a bufio.Writer so we can flush it
+		bw := bufio.NewWriter(w)
+		defer bw.Flush()
+		err = Flow(ctx, src.reader, bw, flow, renderFunc)
+	}
+
+	// If we received a signal, exit with appropriate signal exit code
```
anthonyrisinger, line 408: This definitely needs to be moved to the top of the main function, or it will skip `defers` lower in the stack.

main.go
anthonyrisinger: This is fixed

main.go
anthonyrisinger: This is fixed

main.go
```diff
@@ -417,6 +496,7 @@ func init() {
 	_ = viper.BindPFlag("all", rootCmd.Flags().Lookup("all"))
 
 	viper.SetDefault("style", styles.AutoStyle)
+	viper.SetDefault("flow", 0)
```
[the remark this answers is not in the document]
anthonyrisinger: fixed

main.go
anthonyrisinger: fixed

anthonyrisinger:
I had put this in a review comment, but I don't think it's actionable for this first pass at least, if only because I think it's more important to get the code in the wild first and let the meaning of `--flow=-1024` (if anything) emerge from usage instead; I still think it's a good idea, and suspect there's one or two good meanings already, but there's no pressure to decide right now, and, I think, way more risk of getting it wrong than doing nothing for now:

> I think we could have our cake and eat it too if we interpreted negative sizes as 'aggressive flush first half of double max buffer', i.e. similar to incremental `--flow=N` buffering, but allowing some early opportunistic flushing, maybe we look for two safe boundaries instead of just one or whatever. Ultimately, I think I'm trying to ensure a good user experience for things like long code blocks, but we might need better glamour support or more state tracking so we can indent when the buffering is too small to see the entire code block. This might also enable incremental rendering of things _within_ the code, which is a good thing, but having better support for incremental **anything** in glamour proper would be much nicer, e.g. maybe a few new `Options()` to help manage newlines, large code blocks, large tables, and other large/nested structures we'd like to incrementally render as they come in knowing full well, they are nested within something else. Glamour might return some sort of object that identifies which structure were currently nested in, and then we just continue currying whatever object Glamour returns (it could be a totally black box thing from the consuming API point of view, akin to a cursor).

If e.g. glamour later supported incremental parsing in some way, where callers could pass it a hint of some kind whether as described above or any other way, that could change the game as well, and if it then clashed with whatever we did for negative `--flow` values, it would be obnoxious. So to be clear, I still think it's probably a good idea, but there's more than enough changes already and no reason to shove it through early. The way the code is written right now should be forward compatible with changes we'd want to make in the future.

flow.go
```diff
@@ -0,0 +1,542 @@
...
+			// Header with at least one line after it
+			return true
+		}
+	}
+
+	return false
+}
+
+// findSafeBoundary finds the last safe split point in the buffer
+func (fb *FlowBuffer) findSafeBoundary() int {
```
anthonyrisinger, line 314: We probably want to avoid searching the entire accumulated buffer every time, even though it's kind of an edge case. Since we don't yet consider the internals of tables or code blocks to contain safe boundaries, it's far **more likely** to accumulate larger buffers, but still only really becomes a problem if they become pathologically large. At any rate, keeping track of the already-searched offset is probably straightforward and worth looking at before landing.

anthonyrisinger:
I'll be doing a bit more cleanup tomorrow, but it's very close to its final form; all tests pass in less than a minute now (and should reduce significantly as I port them to Go proper).

flow.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+}
+
+// Process reads markdown from r and streams rendered output via Output callback
+func (fb *FlowBuffer) Process(ctx context.Context, r io.Reader) error {
+	// Channel for read results
+	type readResult struct {
+		data []byte
+		err  error
+	}
```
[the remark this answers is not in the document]
anthonyrisinger: This might go away still before I drop draft, but I don't think it matters much either

main.go
anthonyrisinger: This no longer exists and the line limit works better anyway

main.go
anthonyrisinger: No longer exists

flow.go
```diff
@@ -0,0 +1,542 @@
...
+type FlowRenderFunc func(data []byte) ([]byte, error)
+
+// FlowOutputFunc is called with rendered markdown bytes and a flush flag
+type FlowOutputFunc func(data []byte, flush bool) error
+
+// FlowConfig configures the streaming buffer behavior
+type FlowConfig struct {
+	Window int64          // -1: minimal buffering (aggressive flush), 0: maximum buffering (no flush until EOF), N: bounded at N bytes
+	Render FlowRenderFunc // glamour renderer (byte-based)
+	Output FlowOutputFunc // caller controls output
```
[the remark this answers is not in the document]
anthonyrisinger: This is done

flow.go
anthonyrisinger: This is done

plutack:
This wouldn't work for a pipe command since that blocks right?
My workflow looks something like this..a simple go script that talks to Gemini API then I pipe that to glow.
The last time I tried building this it failed, is it by any chance ready for testing

anthonyrisinger:
@plutack it's been broken on and off, mostly because I keep finding edge cases and learning more about when glow (glamour) outputs white space or does not. I lost a day or so being extremely confused before realizing that a bunch of my tests were using a txt prefix instead of md, which caused vastly different behavior due to automatic fence wrapping by glow. It's been a bit more broken than it was when I first posted because I've been porting everything to go tests, however, the use case that you're describing should 100% be supported, so the moment I feel like it will be stable, which I'm hoping is today, I will ping again.

To be clear it **does** block, but you have the option of choosing how aggressive the `--flow` behavior is. In fully "unbuffered" streaming mode, it will be as close to line buffered as it can possibly be without breaking the markdown structure. Code fencing and tables have special handling to treat them as a singular block across a pretty large window, so you may still see "blocking" for these objects until more advanced inner-content handling is written.

`--flow` is conceptually like a buffer window, but it's a little less rigid than that. It's not a hard limit but rather the "smallest possible chunk" we MUST gather before we will consider it a candidate for a "safe split boundary". Since currently the only known (by me) theoretical difference between streaming markdown and rendering it as one giant blob is resolving reference links at the tail end of the document, e.g. `[See Reference Text][ref-at-eof]`, in a more fluid environment where documents might be streaming for a long time, setting the flow window to twice the size of your largest-expected singular document can **guarantee** resolution of _every_ link.

It was definitely a little trickier than I expected going in, but I also have a pretty high bar for what I expect to work so I will definitely confirm its streams progressively. My own use case is streaming LLM text as well, and I'm implementing this project using a custom agent protocol so it's been a little bit back-and-forth. It's in a pretty good place now, but I'll be pushing all day so it might be best to wait; my goal after landing is to keep hardening it to the point that it can stream inside complex objects as well (I rely on highly-structured colors and capabilities in the terminal to keep me focused and help parse quickly).

plutack:
Thanks for your time and work.

caarlos0:
haven't reviewed the code yet, but did some tests

CleanShot 2025-09-22 at 10 29 23@2x

it seems that small `--flow` values might cause issues? made a shell script that prints some markdown lines with a sleep between them, notice that it didn't syntax highlight or handle the h2 properly.

also, I think having the user decide the buffer size is a bit weird...

caarlos0:
After talking about this with the team, we decided can't move forward with this approach at this time. 
We have another thing in the works that would make this simpler to implement, but we don't have a timeline for it yet.
That said, thanks for bringing this up, and thanks for all the work you put into it.
Again, sorry for not reviewing this earlier, we're a rather small team and sometimes we fail to keep up with everything.

lanmeibuxie:
Thanks for the heads-up! Appreciate the consolidation effort! 👍

anthonyrisinger:
That's fair, and I understand, so no worries at all there ☺️ (I admittedly found `bat` to work well-enough for my original use case FWIW) I did get it working more than a few different ways, but I was never _completely_ satisfied, even in this moment, and none of the methods ever felt particularly 'natural'. As you might be alluding to already, it'd be _much_ easier if chunking/continuation/??? support landed in glamour (or elsewhere) instead; everything I was doing would start fairly clean but if I wanted to handle even moderately-complex documents, compromises came in to play real quick (e.g. streaming the interiors of fences and tables or doing better than line buffer was pretty much impossible without support from glamour). Anyways, the **demand** is real, and glow seems well positioned to solve it! (The version in the PR was definitely broken but I did get line buffered streaming working right, and I can share that if someone ends up asking before it's gone 😁)

anthonyrisinger:
https://gist.github.com/anthonyrisinger/dfcd8a534a974e38160b5d45cc0cc066 This is what I was working on at the time—I was getting annoyed with my agent team flip-flopping back-and-forth in a couple ways, and I wanted a more robust 'truth signal' to help guide them towards preferred completion—might be useful for someone in the future; it quickly confirmed that several of my problems were exactly as nuanced as suspected, e.g. producing white space in situations that felt either ill defined or potentially glamour bugs (esp. around single-item lists and HTML).

> The `flow-tool.go` is a testing and development utility and fake-streaming markdown renderer that generates, processes, and analyzes markdown content through Glamour. It supports flexible markdown primitive generation with customizable repetition and spacing, simulates various streaming modes (unbuffered, buffered, windowed) with artificial delays to test flow control, and provides multiple analysis capabilities including hex dumps, diff comparisons, and whitespace metrics. The tool features a deadman switch mechanism that prevents hangs during stdin processing, making it useful for stress-testing and validating the consistency of streaming implementations across different flow modes and edge cases.

pashpashpash:
> After talking about this with the team, we decided can't move forward with this approach at this time.
> We have another thing in the works that would make this simpler to implement, but we don't have a timeline for it yet.
> That said, thanks for bringing this up, and thanks for all the work you put into it.
> Again, sorry for not reviewing this earlier, we're a rather small team and sometimes we fail to keep up with everything.

@caarlos0 heya, is it possible to share this plan you have in mind for the other thing in the works, the cline team can help out and make a PR. for context, we're working on a cline CLI built in go and would love to use this for rendering streaming markdown.

[https://github.com/cline/cline/pull/6676](https://github.com/cline/cline/pull/6676/files)

PR #867, opened 2026-01-01
fix: Line breaks aren't properly ignored in lists

### LINKED ISSUE #858: Line breaks aren't properly ignored in lists
While line breaks are properly ignored in body text (by the TUI), they aren't ignored when they appear in lists of any sort.

E,g., consider the following markdown:

```markdown
This doesn't have
a line break in it.

This does have\
a line break in it.

Ordered list:
1. This shouldn't have
   a line break
2. This should have\
   a line break

Unordered list:
- This shouldn't have
  a line break
- This should have\
  a line break

Definition list
: This shouldn't have
  a line break
: This should have\
  a line break
```

When rendered by the TUI (with `preserve-new-lines` as false, by default), the first two paragraphs of text render correctly. I.e., the first paragraph is rendered as a single line of text, and the second paragraph contains an explicit line break (I use \\ in the sample markdown for clarity, but it works with two spaces as well).

However, in all of the list examples, both list items contain explicit line breaks. If line breaks are ignored (i.e., not preserved), then the first list item of each list should be rendered on a single line (then reflowed as necessary).

majiayu000:
Fixes #858

## Changes
- Make `WithPreservedNewLines()` conditional based on the `--preserve-new-lines` flag
- Default behavior now correctly treats soft line breaks as spaces in lists

Previously, `glamour.WithPreservedNewLines()` was always applied, causing all line breaks (including soft breaks in lists) to be preserved. Now it respects the CLI flag.

### DIFF
--- a/main.go
+++ b/main.go
@@ -289,13 +289,16 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
 	isCode := !utils.IsMarkdownFile(src.URL)
 
 	// initialize glamour
-	r, err := glamour.NewTermRenderer(
+	options := []glamour.TermRendererOption{
 		glamour.WithColorProfile(lipgloss.ColorProfile()),
 		utils.GlamourStyle(style, isCode),
 		glamour.WithWordWrap(int(width)), //nolint:gosec
 		glamour.WithBaseURL(baseURL),
-		glamour.WithPreservedNewLines(),
-	)
+	}
+	if preserveNewLines {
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

boydfields:
Hello. The binaries have not been updated to correct this fault?

Solsticely:
I'd love for this to be merged upstream! The issue still remains and if I'm correct I believe this can still be merged.

Issue #441, opened 2023-01-16
Feature request: follow relative links in TUI mode


MageJohn:
Often a directory of markdown files will have relative links to each other. It would be nice if in TUI mode it were possible to follow these links directly. 

Note this is seperate from OSC8 hyperlink support (#237). That hands off following the link to the terminal, but in this case it would be useful if Glow were the one following the link so that it opens in the TUI.

To select the links, I imagine pressing tab/shift-tab to move between links and then pressing enter to follow the link. If mouse support is enabled, then it would be awesome to be able to use the mouse as well.

KaviiSuri:
Would love to pick this up if it's available!

carloswm85:
`dart.md` file is in the same folder as `flutter.md`, which contains a link to the former.

![WindowsTerminal_ZKnRaDqdww](https://github.com/charmbracelet/glow/assets/29048235/91c72782-076e-4510-a5f9-fda8346ca658)

What you are saying is that that `./dart.md` link should open the `dart` file? Hyperlinks are properly open when holding `ctrl + left click`, and opening the `dart` file should work the same way.

franalbani:
This would be really nice as it would allow to make a little interactive navigator for a network of markdown files.

askoropadsky:
I'd love this feature too so I can navigate between multiple linked markdown files without taking hands off keyboard!

richcarl:
Keyboard navigation of (local) links was the first thing I expected would Just Work (TM) beyond basic formatting and paging when I picked up Glow.

db7:
Keyboard navigation of links would be awesome, I now think this shouldn't be the task of glow, but rather of the $PAGER. The pager should allow for link selection and open the $BROWSER program with the link. The browser can then decide what to do with the link. For example, if it is a local markdown file in contrast to a https url, then the browser could execv a child glow process to open that file.

What do you think about this design?

Also note, that it might be possible to actually wrap existing pagers to provide such redirection functionality without changing the pager implementations. Some acrobatics with pty should do the job.

djmcgreal-cc:
I think links within the existing file would be easier and better than nothing.

aleexharris:
I would love to see this feature added!

EDIT: I'm going to have a look at this, but I think it requires quite a bit of change to the overall codebase...

aleexharris:
I've opened a new feature request (as per CONTRIBUTING.md) for this here: https://github.com/charmbracelet/glow/discussions/884

[...]

PR #883, opened 2026-02-15
Follow relative markdown links in the TUI pager (#441)

aleexharris:
Addresses long-standing issue #441

New discussion thread https://github.com/charmbracelet/glow/discussions/884

Add TUI pager support for following relative links to other local markdown files via:
- Tab/Shift+Tab to select
- Enter to follow links
- Backspace to go back to parent

Only follow relative links that resolve to an existing regular file within the directory Glow was opened in. Reject absolute paths, UNC/drive paths, external URLs, and root-escape attempts. Added tests for this as it is a safety concern.

PR template:

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).

### DIFF
--- a/ui/keys.go
+++ b/ui/keys.go
@@ -1,6 +1,9 @@
 package ui
 
 const (
-	keyEnter = "enter"
-	keyEsc   = "esc"
+	keyTab       = "tab"
+	keyShiftTab  = "shift+tab"
+	keyEnter     = "enter"
+	keyBackspace = "backspace"
+	keyEsc       = "esc"
 )
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -89,6 +89,11 @@ const (
 	pagerStateStatusMessage
 )
 
+type navEntry struct {
+	Path    string
+	YOffset int
+}
+
 type pagerModel struct {
 	common   *commonModel
 	viewport viewport.Model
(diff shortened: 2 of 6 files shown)


opsec-ai:
#Nice. Thanks for addressing this. As a drive-by tester, I did a `git clone` and applied this PR to it.

Some issues.

* Links are discarded when loading markdown from the command line, e.g. `./glow -t test.md` says "No followable links detected" in spite of there being valid links in the file.
* This was caused by `newModel()` function and all its methods (`Init`,`Update`, `View`) using **value receivers** instead of **pointer receivers** for the `model` struct.
* changes to `m.pager`, `m.stash`, `m.state`, `m.fatalErr` inside these methods are discarded.

Credits due. I used `crush` agent with gemma4 and qwen-3.6 to analyze the project and pinpoint what needs to be changed. Some changes are outlined in this patch:
[link_update.patch](https://github.com/user-attachments/files/28308271/link_update.patch) The rest are staged here: https://github.com/opsec-ai/glow/tree/md_link_fix_pr

| Change | Detail |
|--------|--------|
| `newModel()` return type | `tea.Model` → `*model` (returns pointer directly) |
| `newModel()` returns | `return m` → `return &m` (all 4 return sites) |
| `Init()` receiver | `func (m model)` → `func (m *model)` |
| `Update()` receiver | `func (m model)` → `func (m *model)` |
| `View()` receiver | `func (m model)` → `func (m *model)` |
| `findNextLocalFile()` receiver | `func (m model)` → `func (m *model)` |

### Additional fix in `Init()`

The `stateShowDocument` case now distinguishes between two sources of document content:

- **`localPath != ""`** — reads the file from disk (for files loaded directly or via
stash navigation)
- **`localPath == ""`** — uses the in-memory `Body` (for content passed via stdin or
direct string)

This ensures the document body is correctly populated regardless of how the file was
loaded, which is essential for link extraction to work in all modes.

My latest updates on this subject are at https://github.com/opsec-ai/glow/tree/md_link_fix_pr including a fix for link highlighting. (When pressing Tab, your code highlights the links, but the highlighting does not show up because the viewport wasn't being sync'd.)

PR #899, opened 2026-03-18
Fix TUI not wrapping long lines on initial render

mfdutra:
## Summary

- When opening a markdown file with `-t` (TUI mode), long lines were not wrapped until pressing `r` to refresh
- Root cause: `Init()` called `renderWithGlamour()` before `WindowSizeMsg` arrived, so glamour rendered with `WithWordWrap(0)`
- Fix defers rendering by storing the markdown body until the first `WindowSizeMsg` sets the viewport dimensions, then lets the pager's existing resize handler render with the correct width

## Test plan

- [ ] Open a markdown file with long lines using `glow -t file.md` and verify lines wrap correctly on first render
- [ ] Resize the terminal window and verify content re-wraps correctly
- [ ] Press `r` to refresh and verify it still works
- [ ] Pipe content via stdin (`cat file.md | glow -t`) and verify it renders correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -95,6 +95,9 @@ type commonModel struct {
 	cwd    string
 	width  int
 	height int
+
+	// pendingBody holds markdown content to render once the window size is known.
+	pendingBody *string
 }
 
 type model struct {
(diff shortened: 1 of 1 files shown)


coreyward:
Related: #878 (which already has two other PRs)

PR #938, opened 2026-04-27
TUI: correct redraw on resize and smart word-wrap at terminal width

HollowPhobos:
- [X] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).

## Description

Two related TUI rendering fixes for the pager:

### Fix 1: Correct redraw on terminal resize

Previously, resizing the terminal could result in a blank or stale pager because:
- `contentRenderedMsg` was a plain `string` type, so the original markdown body was lost after the initial render.
- On `WindowSizeMsg`, the code returned early when the body was empty and did not call `viewport.Sync()` in high-performance mode.

**Changes:**
- `contentRenderedMsg` is now a struct carrying both the rendered `content` string and the original `body` (markdown source), which is cached back onto `currentDocument.Body` after rendering.
- On `WindowSizeMsg`, `viewport.Sync()` is called immediately (high-performance mode) and the pager re-renders with the cached body using the new terminal width.
- Initial file render is deferred from `Init()` to after the first `WindowSizeMsg` via `loadLocalMarkdown`, so glamour receives the correct viewport width instead of 0.

### Fix 2: Smart word-wrap at terminal width

Previously, glamour rendering was capped at a hardcoded `GlamourMaxWidth` (120 chars), meaning long lines on wide terminals would overflow the viewport and wrap mid-word by the terminal emulator.

**Changes:**
- Removed the `min(GlamourMaxWidth, viewport.Width)` cap; glamour now renders to the full `viewport.Width`, which is already dynamic.
- `GlamourMaxWidth` is used as a fallback only when `viewport.Width` is 0 (before the first `WindowSizeMsg`).
- Added `softWrapLine`: an ANSI-aware post-render wrapper that searches backwards up to 10 visible characters from the margin for a whitespace or punctuation break point, with a full-line whitespace fallback and a hard-break last resort — preventing mid-word splits regardless of terminal width.

## Steps to reproduce / before & after

**Resize regression:**
1. Open a markdown file with `glow somefile.md`
2. Resize the terminal window
3. **Before:** pager goes blank or retains stale content at the old width
4. **After:** pager re-renders correctly at the new width

**Word-wrap:**
1. Open a markdown file containing long lines on a terminal wider than 120 columns
2. **Before:** lines overflow and are wrapped mid-word by the terminal
3. **After:** lines wrap cleanly at word/punctuation boundaries within the terminal width

### DIFF
--- a/main.go
+++ b/main.go
@@ -199,9 +199,7 @@ func validateOptions(cmd *cobra.Command) error {
 				width = uint(w) //nolint:gosec
 			}
 
-			if width > 120 {
-				width = 120
-			}
+
 		}
 		if width == 0 {
 			width = 80
--- a/ui/pager.go
+++ b/ui/pager.go
@@ -78,8 +78,11 @@ var (
 )
 
 type (
-	contentRenderedMsg string
-	reloadMsg          struct{}
+	contentRenderedMsg struct {
+		content string
+		body    string // original markdown body, cached for re-render on resize
+	}
+	reloadMsg struct{}
 )
 
 type pagerState int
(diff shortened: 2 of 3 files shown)


HollowPhobos:
closed for being ignored

PR #966, opened 2026-05-30
fix: use /dev/tty for width autodetect; default config width: 0

HeikoSchlittermann:
Two related width-autodetect fixes.

**1. Prefer `/dev/tty` over `os.Stdout` for terminal-size detection.**

`term.GetSize(os.Stdout.Fd())` fails whenever stdout isn't a TTY — e.g. `glow x.md | less`, `glow x.md > out`, or some `glow <x.md` setups — and glow silently falls back to 80 columns. `/dev/tty` refers to the controlling terminal regardless of how the standard fds are wired, so it works in all those cases. Falls back to the original `term.GetSize(os.Stdout.Fd())` path when `/dev/tty` can't be opened, so behaviour is a strict superset.

**2. Default config now ships `width: 0` instead of `width: 80`.**

`glow config` writes a hard-coded YAML template on first run with `width: 80`, which silently overrides the documented `0 = autodetect` default (`viper.SetDefault("width", 0)` in main.go). The auto-generated config is therefore strictly more restrictive than no config at all. Aligning the template with `SetDefault` restores the documented behaviour.

Existing users with the old config still get 80 until they edit it once; new users get autodetect out of the box.

### DIFF
--- a/config_cmd.go
+++ b/config_cmd.go
@@ -19,8 +19,8 @@ style: "auto"
 mouse: false
 # use pager to display markdown
 pager: false
-# word-wrap at width
-width: 80
+# word-wrap at width (0 = autodetect terminal width)
+width: 0
 # show all files, including hidden and ignored.
 all: false
 `
--- a/glow_test.go
+++ b/glow_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"testing"
 )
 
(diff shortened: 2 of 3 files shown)


HeikoSchlittermann:
Related: #952 tackles a third aspect of the same area — removing the 120-column cap on the auto-detected width. The two PRs are orthogonal (this one fixes *which fd* is queried and the default-config template; #952 fixes the *cap* applied afterwards) and could be merged independently or together.

PR #985, opened 2026-06-25
fix(cli): respect preserve-new-lines flag in CLI renderer

syf2211:
## Summary
CLI mode always passed `glamour.WithPreservedNewLines()`, ignoring the `-n` flag and config default.

## Motivation
TUI mode already respects `preserveNewLines` in `ui/pager.go`, but `executeCLI` forced preserved newlines for every render. That made CLI output differ from TUI and contributed to glow #858 list rendering complaints.

## Changes
- Build CLI glamour options conditionally, matching TUI behavior
- Add `TestCLIPreserveNewLinesFlag` to verify the default path collapses soft breaks

## Tests
- `go test ./... -count=1` — pass

## Notes
Full fix for list soft breaks inside list items requires the companion glamour change in charmbracelet/glamour#580 and a follow-up dependency bump once released.

Related to #858

### DIFF
--- a/glow_test.go
+++ b/glow_test.go
@@ -1,10 +1,22 @@
 package main
 
 import (
+	"bytes"
+	"io"
+	"regexp"
+	"strings"
+	"sync"
 	"testing"
+
+	"github.com/spf13/viper"
 )
 
+var cliTestMu sync.Mutex
+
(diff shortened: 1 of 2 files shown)


yndajas:
Encountered this issue and having to unwrap the lines with pandoc before passing  to glow

PR #1002, opened 2026-08-03
chore: bump bubbletea to v2

### LINKED ISSUE #554: glow shuffles lines randomly after scrolling using u/d buttons or mouse
**Describe the bug**
After some scrolling lines start to shuffle randomly.

**Setup**
 - OS: ubuntu 22.04.3; i3wm
 - Shell: bash 
 - Terminal Emulator: kitty 0.30.0
 - Terminal Multiplexer: none
 - Locale: en_US.UTF-8, ru_RU.UTF-8
 - Go version: go1.20.4 linux/amd64

**To Reproduce**
1. Start glow not in pager mode: `glow`
2. Open any document
3. Scroll randomly (with a mouse or by using `u/d`)
4. See some lines shuffle

**Expected behavior**
Lines shouldn't be shuffled.

**Screenshots**
The original file is on the right. It is clear that glow messed some lines in the last paragraph (on the left):
Screenshot from 2024-01-08 00-16-44
Other lines shuffle after some scrolling:
Screenshot from 2024-01-08 00-17-45
It is especially sad when a code section is messed up: 
Screenshot from 2024-01-08 00-18-19

**Additional context**
If a file is opened in pager mode (`glow -p filename`) then scrolling works fine and nothing shuffles.

Scrolling by using `b/f/j/k/g/G` seems to work fine.
Changing `pager` variable in `glow.yml` doesn't help.
I built glow from source. The latest commit in my local repo is `54dd62a2a43d36e10d075b2b3fe9b0f637f2e2b9`.

### LINKED ISSUE #595: paragraph/heading/list word-wrap never breaks unbroken CJK text
### Description

> Note: I'm not confident in my written English, so this issue was translated with the help of an AI assistant. Please excuse any unnatural phrasing.

### Symptom

Even when a width is specified via `WordWrap`, paragraphs, headings, and list items written in a language without ASCII spaces (such as Japanese) are **never wrapped at all**. The output ignores the specified width and renders everything on a single line.

### Steps to reproduce

```go
r, _ := glamour.NewTermRenderer(
      glamour.WithStandardStyle("dark"),
      glamour.WithWordWrap(40),
)
out, _ := r.Render("これは日本語のように単語間にスペースを含まない言語で書かれた非常に長い段落です。")
```

out is not wrapped despite WordWrap: 40, and stays on a single line:

これは日本語のように単語間にスペースを含まない言語で書かれた非常に長い段落です。

Root cause

ansi/paragraph.go and ansi/heading.go use github.com/muesli/reflow/wordwrap. ansi/blockelement.go (which margin-bearing blocks such as lists and blockquotes go through) uses charmbracelet/x/ansi.Wordwrap.

Both are designed to only break at word boundaries, with no hard-wrap fallback for a "word" that itself exceeds the given width. In languages like Japanese that don't separate words with spaces, an entire block of text is treated as a single "word," so no line break ever occurs regardless of the configured width.

By contrast, ansi/table.go delegates table rendering to lipgloss/table, which internally uses ansi.Wrap — a function that does have a hard-wrap fallback for oversized words — so tables are not affected. I verified this behavioral difference between ansi.Wrap and ansi.Wordwrap directly against real Japanese text:

- ansi.Wrap: correctly wraps within the specified width
- ansi.Wordwrap / muesli/reflow/wordwrap: neither wraps at all, and both produce lines far exceeding the specified width (in one real-world example, a width of 100 was requested but lines of width 314–372 were produced)

Note: already fixed on the v2 line

On the master branch (the v2 line), paragraph.go has already moved to lipgloss.Wrap (refactor(ansi): drop reflow and use cellbuf.Wrap), so this issue does not occur there. However, it has not been backported to the currently tagged release line (v0.10.x — the version fetched via go get, and also the version charmbracelet/glow currently depends on).

Environment

- glamour v0.10.0
- Example of a downstream consumer affected: glow v2.1.1 (rendering Markdown containing CJK text, e.g. some-command | glow -, breaks paragraphs, headings, and list items)

Fix

I plan to submit a separate PR that switches paragraph.go, heading.go, and blockelement.go to use ansi.Wrap.

### Version

v0.10.0

### Environment

Ubuntu 24.04.4 LTS (WSL2, Linux 6.6.114.1-microsoft-standard-WSL2, x86_64), go1.26.3, glamour v0.10.0, glow v2.1.1

### LINKED ISSUE #999: Double-rendering after closing editor
**Describe the bug**
Hi!
I am using `glow` in my terminal. I opened a Markdown file as usual, then found out I can edit it directly from `glow` using `nano`, by just pressing `e`. I did it, and I pressed `^O ^X` to save the file and exit. But then, although the file was saved correctly (all its contents were overriden by the new content I wrote in Nano), Glow showed it twice - the original content, a line break, then the new file contents. I know that my description isn't really understandable, so please open any Markdown in `glow`, press `e` to edit it in your favorite editor, close the editor and see the results.

**Setup**
 - OS [e.g. Ubuntu, macOS] macOS Tahoe 26.6
 - Shell [e.g. zsh, fish] zsh
 - Terminal Emulator [e.g. kitty, iterm] Apple Terminal.app
 - Terminal Multiplexer [e.g. tmux]
 - Locale [e.g. en_US.UTF-8, zh_CN.UTF-8, etc.] en_US.UTF-8

**To Reproduce**
1. Go to any Markdown file
2. Click on the keyboard button `e`.
3. Edit the file noticably (even rewrite its contents), save it and close the editor

**Source Code**
Please include source code if needed to reproduce the behavior. 

**Expected behavior**
I expected that `glow` shows ONLY the new version of the file, not both the old and new.

**Screenshots**

Image

**Additional context**

andrinoff:
## What?

Bumps bubbletea to v2

## Why?

Fixes #554 
Fixes #999 as `HighPerformancePager` is depreciated in v2

Fixes charmbracelet/glamour#595


andrinoff:
additionally, upgrading glamour to v2, due to charmbracelet/glamour#595

ui/ui.go
```diff
@@ -120,21 +114,18 @@ func (m *model) unloadDocument() []tea.Cmd {
...
-
 	if !m.stash.shouldSpin() {
 		batch = append(batch, m.stash.spinner.Tick)
 	}
 	return batch
 }
 
 func newModel(cfg Config, content string) tea.Model {
+	initStyles()
 	initSections()
```
meowgorithm: These should probably set members on the model rather than global vars, which means these functions should be methods the model. Less global state floating around.

ui/ui.go
```diff
@@ -120,21 +114,18 @@ func (m *model) unloadDocument() []tea.Cmd {
...
 
-	if cfg.GlamourStyle == styles.AutoStyle {
-		if te.HasDarkBackground() {
+	if cfg.GlamourStyle == "auto" {
+		if lipgloss.HasDarkBackground(os.Stdin, os.Stdout) {
 			cfg.GlamourStyle = styles.DarkStyle
 		} else {
 			cfg.GlamourStyle = styles.LightStyle
```
meowgorithm: This code can go. We don't need to detect this here anymore because Bubble Tea is now doing background color detection asynchronously elsewhere in this PR (which is the way to go).

ui/ui.go
```diff
@@ -205,15 +196,18 @@ func (m model) Init() tea.Cmd {
...
 		}
 	}
 
 	var cmds []tea.Cmd
 
 	switch msg := msg.(type) {
-	case tea.KeyMsg:
+	case tea.BackgroundColorMsg:
+		lightDark = lipgloss.LightDark(msg.IsDark())
+		initStyles()
```
meowgorithm: Just to make it super obvious per my comment above, this is where the background color is being correctly detected.

ui/ui.go
```diff
@@ -324,17 +318,23 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 	}
 
-	switch m.state { //nolint:exhaustive
-	case stateShowDocument:
-		return m.pager.View()
-	default:
-		return m.stash.view()
+	v := tea.NewView(content)
+	v.AltScreen = true
+	if m.common.cfg.EnableMouse {
```
meowgorithm, line 318: For a follow up: remove this config and add text selection. This setting exists because, at the time, you had to choose between being able to select text and using the mouse.

utils/utils.go
```diff
@@ -72,8 +72,8 @@ func IsMarkdownFile(filename string) bool {
 // GlamourStyle returns a glamour.TermRendererOption based on the given style.
 func GlamourStyle(style string, isCode bool) glamour.TermRendererOption {
 	if !isCode {
-		if style == styles.AutoStyle {
-			return glamour.WithAutoStyle()
+		if style == "auto" {
+			return glamour.WithStandardStyle("dark")
```
meowgorithm, line 76: Make sure we're not using this function in Bubble Tea and only in the CLI (single Bubble Tea handles background color detection).

ui/styles.go
```diff
@@ -1,46 +1,201 @@
...
+			Foreground(lipgloss.Color("#ECFD65")).
+			Background(fuchsia).
+			Bold(true)
+
+	stashSpinnerStyle = lipgloss.NewStyle().
+				Foreground(gray)
+	stashInputPromptStyle = lipgloss.NewStyle().
+				Foreground(yellowGreen).
+				MarginRight(1)
+}
```
meowgorithm: As mentioned elsewhere, I'd move these onto the `Model`, probably as a `Styles` struct, so they're not all top level.

ui/styles.go
```diff
@@ -1,46 +1,201 @@
...
+	cream          color.Color
+	yellowGreen    color.Color
+	fuchsia        color.Color
+	dimFuchsia     color.Color
+	dullFuchsia    color.Color
+	dimDullFuchsia color.Color
+	green          color.Color
+	red            color.Color
+	semiDimGreen   color.Color
+	dimGreen       color.Color
```
meowgorithm: Worth checking if we still need these.

ui/styles.go
```diff
@@ -1,46 +1,201 @@
...
+	lineNumberStyle                func(...string) string
+)
+
+// Stash styles.
+var (
+	dividerDot            lipgloss.Style
+	dividerBar            lipgloss.Style
+	logoStyle             lipgloss.Style
+	stashSpinnerStyle     lipgloss.Style
+	stashInputPromptStyle lipgloss.Style
```
meowgorithm: Also worth checking if we still need these, per the comment above.

meowgorithm (changes requested):
Working _great_! Only comments here are around background color detection and reducing globals. Nice work!!

andrinoff:
@meowgorithm refactored! 

good to merge?

meowgorithm (approved):

PR #1015, opened 2026-08-13
feat: hyperlinks

andrinoff:
## What?

Detect OSC 8 hyperlink support from the environment (`utils.SupportsHyperlinks`) and pass `glamour.WithHyperlinkMode(ansi.HyperlinkModeInline)` when supported, in both the pager and CLI paths. Uses a pinned glamour version (from the PR)


sananiduo:
I added https://github.com/charmbracelet/glow/pull/1016 to also make it work with a config option to be able to support combinations like I'm using (tmux+iterm2) https://github.com/charmbracelet/glow/pull/1016

sananiduo:
@andrinoff everything alright? :)

andreynering (approved):

main.go
```diff
@@ -290,12 +291,16 @@ func executeCLI(cmd *cobra.Command, src *source, w io.Writer) error {
...
 	// initialize glamour
-	r, err := glamour.NewTermRenderer(
+	options := []glamour.TermRendererOption{
 		utils.GlamourStyle(style, isCode),
 		glamour.WithWordWrap(int(width)), //nolint:gosec
 		glamour.WithBaseURL(baseURL),
 		glamour.WithPreservedNewLines(),
-	)
+	}
+	if utils.SupportsHyperlinks(os.Environ()) {
```
taciturnaxolotl, line 300: this feels like a bad config name since we still omit OSC 8 hyperlinks; its closer to a hyperlink display format

utils/utils.go
```diff
@@ -111,3 +111,39 @@ func GlamourStyle(style string, isCode bool) glamour.TermRendererOption {
...
+		return ""
+	}
+
+	// Explicit opt-out.
+	if get("TERM_PROGRAM") == "Apple_Terminal" {
+		return false
+	}
+	// Terminals known to support OSC 8 hyperlinks.
+	switch get("TERM_PROGRAM") {
+	case "iTerm.app", "WezTerm", "kitty", "alacritty", "Hyper", "ghostty", "vscode", "Tabby", " Rio":
```
taciturnaxolotl: rio sets `TERM_PROGRAM=rio` so this won't match; alacritty only sets ALACRITTY_WINDOW_ID, USER, HOME, WINDOWID, and TERM=alacritty

taciturnaxolotl: we should probably also check if its a interactive tty as well

taciturnaxolotl: https://github.com/zkat/supports-hyperlinks/blob/main/src/lib.rs has a bunch of really good checks that may be worth borrowing

PR #1028, opened 2026-09-01
feat: image support

### LINKED ISSUE #211: Feature-request: Display images in terminal
In markdown you can display images using \[name](/path/to/image).

It would be very cool to be able display them in the terminal with glow. 

It could use many techniques to display images :
- Graphical libraries (sdl) in terminal
- Framebuffer 
- Ascii art as done with bubbletea imgcat?

andrinoff:
## What?

Glow can now display images from markdown documents directly in the terminal, both in the TUI pager and in plain CLI output. It works on terminals that support the kitty graphics protocol (kitty, Ghostty, WezTerm) or sixel; support is detected automatically. You can override detection with `GLOW_IMAGE_PROTOCOL` (`auto`, `kitty`, `sixel`, `none`) or turn images off with `--images=false` / `images: false` in the config file.

In the TUI, images are anchored to the text itself, so they scroll smoothly with the document at no extra cost. Images are automatically shrunk to what the terminal can actually display before being sent, so even large photos appear quickly. Their size can also be capped with `—image-max-rows`, config key `imageMaxRows`, `GLOW_IMAGE_MAX_ROWS` (default 20, 0 for no limit).

## Why?

Closes #211

---
<a href="https://app.blacksmith.sh/charmbracelet/codesmith/glow/pr/1028"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790841109&installation_model_id=432448&pr_number=1028&repository=charmbracelet%2Fglow&return_to=https%3A%2F%2Fgithub.com%2Fcharmbracelet%2Fglow%2Fpull%2F1028&signature=7793c796a7eb13bd250f2a3b46f021e6e4dc409c94098b2755444e522626969c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


smorrisods:
Thanks for your work here! This would be an epic feature I just realized could be very useful 😄. Subscribing 🔔

taciturnaxolotl:
this is great! one potential problem is that we don't check the size of an image before loading it so glow will happily oom itself and load a 500 megapixel image; the same happens for remote images as well

imageCache, imageConfigCache, remoteImageCache, sequenceCache are never evicted; it might not be a bad idea to stick them in the TermRenderer and evict them when switching documents

taciturnaxolotl:
any images not http(s):// are treated as local files which breaks ftp: or data: uris

for kitty images all images are transmitted as local files which will break over ssh

taciturnaxolotl:
we may want to consider not loading external images by default as they could contain a tracking pixel + use unwanted bandwidth

taciturnaxolotl:
oops these comments should be on the glamour pr probably

taciturnaxolotl:
absolute paths for images don't resolve correctly; FileBaseURL also breaks when you pass it relative inputs which currently isn't a problem but could be a potential problem latter

ui/pager.go
```diff
@@ -182,9 +195,26 @@ func (m pagerModel) update(msg tea.Msg) (pagerModel, tea.Cmd) {
...
+			// this model sees it, so stash the content and wait for it.
+			m.pendingContent = msg.content
+			return m, tea.Raw(strings.Join(msg.graphics, ""))
+		}
+
+		m.setContent(msg.content)
 		cmds = append(cmds, m.watchFile)
 
+	// The graphics sequences for the pending content have been written
+	case tea.RawMsg:
```
taciturnaxolotl, line 211: this really doesn't feel great; I would use a custom message type

